PR #1, opened 2017-07-13
Add travis testing and deploy

fsih:


.eslintignore
```diff
@@ -1,2 +1,3 @@
 node_modules/*
-build/*
\ No newline at end of file
+build/*
+playground/
```
rschamp: Nit: Add a newline to the end of the file

.gitignore
```diff
@@ -3,4 +3,8 @@
 
 # NPM
 /node_modules
-npm-*
\ No newline at end of file
+npm-*
+
+# Build
+playground/*
+build/*
```
rschamp: Nit: newline

package.json
```diff
@@ -1,22 +1,23 @@
 {
   "name": "scratch-blobs",
   "version": "0.1.0",
-  "description": "GraphicaL User Interface for the Scratch 3.0 paint editor, which is used to make and edit sprites for use in projects.",
+  "description": "Graphical User Interface for the Scratch 3.0 paint editor, which is used to make and edit sprites for use in projects.",
```
rschamp, line 4: FYI that capitalized L is the remnant of when we couldn't decide whether to call it the GLUI or GUI.  I think it might be good to call out in this description that it provides a React component.

paulkaplan: GLUI.... ?!

.eslintignore
paulkaplan: tip: there are usually editor plugins to force this

src/index.js
```diff
@@ -1,3 +1,3 @@
 import PaintEditorComponent from './components/paint-editor.jsx';
 
-export {PaintEditorComponent as default};
\ No newline at end of file
+export {PaintEditorComponent as default};
```
paulkaplan: is `export default PaintEditorComponent` more idiomatic @rschamp? I'm still an export NOOB.

package.json
rschamp: Yeah it's the part that glues the other parts together. It was cute but ultimately suffered the cold blade of practicality.

src/index.js
rschamp: Yeah... I think this is left over from when things were moving around for the playground? This could just be a one line file, `export default PaintEditorComponent from './components/paint-editor.jsx';`

PR #2, opened 2017-07-25
Add PaperCanvas component

fsih:
Adds a canvas that paper.js hooks up to (and draws to with some dummy code), and Redux state that keeps track of the tool. Pressing 'b' and 'e' changes the tool state, which does nothing yet.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,4 +1,3 @@
 node_modules/*
 dist/*
 playground/
-webpack.config.js
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -3,12 +3,12 @@ const path = require('path');
 const webpack = require('webpack');
 
 // Plugins
-var HtmlWebpackPlugin = require('html-webpack-plugin');
+const HtmlWebpackPlugin = require('html-webpack-plugin');
 
 // PostCss
-var autoprefixer = require('autoprefixer');
-var postcssVars = require('postcss-simple-vars');
-var postcssImport = require('postcss-import');
+const autoprefixer = require('autoprefixer');
+const postcssVars = require('postcss-simple-vars');
+const postcssImport = require('postcss-import');
 
 const base = {
     devtool: 'cheap-module-source-map',


src/containers/paint-editor.jsx
```diff
@@ -0,0 +1,48 @@
...
+    onKeyPress: e => {
+        if (e.key === 'e') {
+            dispatch(tools.changeTool(ToolTypes.ERASER));
+        } else if (e.key === 'b') {
+            dispatch(tools.changeTool(ToolTypes.BRUSH));
+        }
+    }
+});
+
+module.exports = connect(
```
paulkaplan: We changed the syntax in gui to use `export`, I think you may want to do that here as well. here is the PR that included the eslint rules https://github.com/LLK/scratch-gui/pull/552/files

src/containers/paint-editor.jsx
```diff
@@ -0,0 +1,48 @@
...
+import ToolTypes from '../tools/tool-types.js';
+import {connect} from 'react-redux';
+
+class PaintEditor extends React.Component {
+    componentDidMount () {
+        const onKeyPress = this.props.onKeyPress;
+        document.onkeydown = function (e) {
+            e = e || window.event;
+            onKeyPress(e);
+        };
```
rschamp: A better convention for this listener is to attach the handler with `attachEventListener`, as in the GUI: https://github.com/LLK/scratch-gui/blob/747924ae2ad5bc480e5b84e95b885856561429b7/src/containers/green-flag.jsx#L24

And don't forget to remove the listener on `componentDidUnmount`, also in that example

src/containers/paint-editor.jsx
```diff
@@ -0,0 +1,48 @@
...
+    }
+    render () {
+        return (
+            <PaintEditorComponent
+                tool={this.props.tool}
+            />
```
rschamp: Convention in scratch-gui, which might be useful here:
```jsx
const {
    onKeyPress, // and any other props consumed by this container
    ...props
} = this.props;
return (
    <PaintEditorComponent ...props />
);
```

src/containers/paper-canvas.jsx
```diff
@@ -0,0 +1,47 @@
...
+import ToolTypes from '../tools/tool-types.js';
+
+class PaperCanvas extends React.Component {
+    constructor (props) {
+        super(props);
+        this.state = {
+        };
+    }
+    componentDidMount () {
+        paper.setup('paper-canvas');
```
rschamp: Can Paper set up using a DOM element instead of an id? If so, a better way to do this would be to use a ref to pass it the canvas component and avoid referencing by id. https://facebook.github.io/react/docs/refs-and-the-dom.html

src/containers/paper-canvas.jsx
```diff
@@ -0,0 +1,47 @@
...
+        // Draw the view now:
+        paper.view.draw();
+    }
+    componentWillReceiveProps (nextProps) {
+        if (nextProps.tool !== this.props.tool && nextProps.tool instanceof ToolTypes) {
+            // TODO switch tool
+        }
+    }
+    componentWillUnmount () {
+    }
```
rschamp: Anything we need to do here to unload Paper?

src/containers/paper-canvas.jsx
```diff
@@ -0,0 +1,47 @@
...
+        // Move to start and draw a line from there
+        path.moveTo(start);
+        // Note that the plus operator on Point objects does not work
+        // in JavaScript. Instead, we need to call the add() function:
+        path.lineTo(start.add([200, -50]));
+        // Draw the view now:
+        paper.view.draw();
+    }
+    componentWillReceiveProps (nextProps) {
+        if (nextProps.tool !== this.props.tool && nextProps.tool instanceof ToolTypes) {
```
rschamp: This validation could happen with PropTypes.  These conflict too: is it a string or a `ToolType`?

src/playground/playground.jsx
```diff
@@ -1,9 +1,21 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import PaintEditor from '..';
+import {Provider} from 'react-redux';
+import {createStore, applyMiddleware} from 'redux';
+import throttle from 'redux-throttle';
```
rschamp: You probably won't need the `throttle` middleware.

src/playground/playground.jsx
```diff
@@ -1,9 +1,21 @@
...
 document.body.appendChild(appTarget);
-ReactDOM.render(
-    <PaintEditor />,
-    appTarget);
+const store = applyMiddleware(
+    throttle(300, {leading: true, trailing: true})
+)(createStore)(
+    reducer,
+    window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__()
+);
```
rschamp: In the case of not using `throttle`, this can be simplified to
```javascript
const store = createStore(
    reducer,
    window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__()
);
```

src/reducers/tools.js
```diff
@@ -0,0 +1,32 @@
...
+};
+
+// Action creators ==================================
+reducer.changeTool = function (tool) {
+    return {
+        type: CHANGE_TOOL,
+        tool: tool,
+        meta: {
+            throttle: 30
+        }
```
rschamp: This `meta` section can be removed if you're not throttling.

src/tools/tool-types.js
```diff
@@ -0,0 +1,12 @@
...
+        this.name = name;
+    }
+    toString () {
+        return `ToolTypes.${this.name}`;
+    }
+}
+ToolTypes.BRUSH = new ToolTypes('BRUSH');
+ToolTypes.ERASER = new ToolTypes('ERASER');
+
+module.exports = ToolTypes;
```
rschamp: In www we use the npm package`keymirror` for this kind of thing (which came out of React).

```javascript
const ToolTypes = keyMirror({
    BRUSH: null,
    ERASER: null
});

if (tool in ToolTypes) { // valid
} else { // invalid
}
```

package.json
```diff
@@ -38,19 +39,23 @@
     "eslint-plugin-react": "^7.0.1",
     "gh-pages": "github:rschamp/gh-pages#publish-branch-to-subfolder",
     "html-webpack-plugin": "2.28.0",
+    "lodash.defaultsdeep": "4.6.0",
     "minilog": "3.1.0",
     "mkdirp": "^0.5.1",
+    "paper": "^0.11.4",
```
rschamp: This should be an exact version since it's used in the application

src/log/log.js
```diff
@@ -0,0 +1,4 @@
+const minilog = require('minilog');
+minilog.enable();
+
+module.exports = minilog('paint-editor');
```
rschamp: By convention the namespace here should be the package, `scratch-paint`.

src/reducers/tools.js
```diff
@@ -0,0 +1,32 @@
...
+const initialState = ToolTypes.BRUSH;
+
+const reducer = function (state, action) {
+    if (typeof state === 'undefined') state = initialState;
+    switch (action.type) {
+    case CHANGE_TOOL:
+        if (action.tool instanceof ToolTypes) {
+            return action.tool;
+        }
+        log.warn(`Warning: Tool type does not exist: ${action.tool}`);
```
rschamp: "Warning: " probably not necessary if you're using `log.warn` :P

src/reducers/tools.js
fsih: Why do we have throttling in the GUI but wouldn't want it in the paint editor?

PR #3, opened 2017-07-28
Add brush and eraser tools

fsih:
Pulls in brush and eraser tools from https://github.com/fsih/papergrapher
Brush size adjustment is bound to mouse scroll, but that is temporary until we add an interface for changing it.

### DIFF
--- a/src/containers/blob/blob.js
+++ b/src/containers/blob/blob.js
@@ -82,9 +82,9 @@ class Blobbiness {
             blob.resizeCursorIfNeeded(event.point);
             if (event.event.button > 0) return;  // only first mouse button
             if (blob.brush === Blobbiness.BROAD) {
-                blob.broadBrushHelper.onBroadMouseDrag(event, blob.options);
+                blob.broadBrushHelper.onBroadMouseDrag(event, blob.tool, blob.options);
             } else if (blob.brush === Blobbiness.SEGMENT) {
-                blob.segmentBrushHelper.onSegmentMouseDrag(event, blob.options);
+                blob.segmentBrushHelper.onSegmentMouseDrag(event, blob.tool, blob.options);
             } else {
                 log.warn(`Brush type does not exist: ${blob.brush}`);
             }
@@ -102,7 +102,7 @@ class Blobbiness {
             if (blob.brush === Blobbiness.BROAD) {
                 lastPath = blob.broadBrushHelper.onBroadMouseUp(event, blob.tool, blob.options);
             } else if (blob.brush === Blobbiness.SEGMENT) {
-                lastPath = blob.segmentBrushHelper.onSegmentMouseUp(event);
+                lastPath = blob.segmentBrushHelper.onSegmentMouseUp(event, blob.tool, blob.options);
             } else {
                 log.warn(`Brush type does not exist: ${blob.brush}`);
             }
--- a/src/containers/blob/broad-brush-helper.js
+++ b/src/containers/blob/broad-brush-helper.js
@@ -30,7 +30,7 @@ class BroadBrushHelper {
         this.lastPoint = this.secondLastPoint = event.point;
     }
     
-    onBroadMouseDrag (event, options) {
+    onBroadMouseDrag (event, tool, options) {
         const step = (event.delta).normalize(options.brushSize / 2);
 
         // Move the first point out away from the drag so that the end of the path is rounded
--- a/src/containers/blob/segment-brush-helper.js
+++ b/src/containers/blob/segment-brush-helper.js
@@ -36,7 +36,7 @@ class SegmentBrushHelper {
         this.lastPoint = event.point;
     }
     
-    onSegmentMouseDrag (event, options) {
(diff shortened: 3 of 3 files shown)


src/reducers/brush-tool.js
```diff
@@ -0,0 +1,28 @@
...
+            return state;
+        }
+        return {brushSize: Math.max(1, action.brushSize)};
+    default:
+        return state;
+    }
+};
+
+// Action creators ==================================
+reducer.changeBrushSize = function (brushSize) {
```
rschamp: With `import`/`export`, we can export the action creators separately, e.g.,
```javascript
const changeBrushSize = ...

export {
    reducer as default,
    changeBrushSize
};

File: src/containers/tools/brush-tool.jsx [hunk 2]
```diff
@@ -0,0 +1,99 @@
...
+            this.activateTool(this.props);
+        }
+    }
+    componentWillReceiveProps (nextProps) {
+        if (nextProps.tool === BrushTool.TOOL_TYPE && this.props.tool !== BrushTool.TOOL_TYPE) {
+            this.activateTool();
+        } else if (nextProps.tool !== BrushTool.TOOL_TYPE && this.props.tool === BrushTool.TOOL_TYPE) {
+            this.deactivateTool();
+        } else if (nextProps.tool === BrushTool.TOOL_TYPE && this.props.tool === BrushTool.TOOL_TYPE) {
+            this.blob.setOptions(nextProps.brushToolState);
```
rschamp: I think you could consolidate some of this logic into `mapStateToProps`, and then have the API within the component just switch based on `activated: true/false`? This might also let you refactor that activation logic into a decorator or higher-order component.

src/containers/tools/brush-tool.jsx
```diff
@@ -0,0 +1,99 @@
...
+        if (event.deltaY < 0) {
+            this.props.changeBrushSize(this.props.brushToolState.brushSize + 1);
+        } else if (event.deltaY > 0 && this.props.brushToolState.brushSize > 1) {
+            this.props.changeBrushSize(this.props.brushToolState.brushSize - 1);
+        }
+        return true;
+    }
+    render () {
+        return (
+            <div>Brush Tool</div>
```
rschamp: It seems like this component wants to be a HOC, since ideally containers shouldn't have any control over presentation, they should just wrap the UI their the container for. Do you think it'd make sense down the line for this to wrap the tool button component?

src/tools/broad-brush-helper.js
```diff
@@ -0,0 +1,111 @@
...
+ * in a flat edge.
+ *
+ * @param {!Tool} tool paper.js mouse object
+ */
+const broadBrushHelper = function (tool) {
+    let lastPoint;
+    let secondLastPoint;
+    let finalPath;
+
+    tool.onBroadMouseDown = function (event) {
```
rschamp: If possible, I would want to avoid mutating function arguments, since this can lead to hard-to-debug behavior, and is harder to test since you can't test its methods without a mock of the mutated component. A decorator should return a new copy of what you pass in with these features added, using a reference to the passed in component.  Is it possible to design this that way?

If not, I think a better name for this function would be `applyBroadBrushFeatures` or something like that?

src/tools/segment-brush-helper.js
```diff
@@ -0,0 +1,85 @@
...
+
+    tool.onSegmentMouseUp = function (event) {
+        if (event.event.button > 0) return;  // only first mouse button
+
+        // TODO: This smoothing tends to cut off large portions of the path! Would like to eventually
+        // add back smoothing, maybe a custom implementation that only applies to a subset of the line?
+
+        // Smooth the path.
+        finalPath.simplify(2);
+        // console.log(finalPath.segments);
```
rschamp: Nit: Remove commented out code

src/containers/tools/brush-tool.jsx
fsih: The brush and eraser modes are going to be closely tied to the brush and eraser buttons, so I decided to leave these in containers. The components will be the buttons.

src/tools/broad-brush-helper.js
fsih: I changed the helpers to classes that Blobbiness has an instance of. It calls each of the functions with a bunch of parameters instead of tool holding them all.

src/modes/broad-brush-helper.js
```diff
@@ -11,101 +12,103 @@ import paper from 'paper';
...
-    };
+        this.finalPath = new paper.Path();
+        stylePath(this.finalPath);
+        this.finalPath.add(event.point);
+        this.lastPoint = this.secondLastPoint = event.point;
+    }
     
-    tool.onBroadMouseDrag = function (event) {
-        const step = (event.delta).normalize(this.options.brushSize / 2);
+    onBroadMouseDrag (event, options) {
```
rschamp: Let's keep the second argument consistent across these functions — the linter won't complain if the second argument isn't used but the third is.

PR #9, opened 2017-09-05
Support import and export of costumes in the paint editor

fsih:
The paint editor takes an optional SVG prop type which it imports, as well as a rotation center point so that it can correctly center the view in the paint editor. Changing the SVG in the props clears the paint editor and draws the new SVG. Drawing in the paint editor calls the updateSvg function passed in via props with the new SVG and rotation center.

### DIFF
--- a/src/containers/paint-editor.jsx
+++ b/src/containers/paint-editor.jsx
@@ -21,9 +21,6 @@ class PaintEditor extends React.Component {
         document.removeEventListener('keydown', this.props.onKeyPress);
     }
     handleUpdateSvg () {
-        if (!this.props.onUpdateSvg) {
-            return;
-        }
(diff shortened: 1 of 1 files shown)


src/containers/paint-editor.jsx
```diff
@@ -4,23 +4,53 @@ import PaintEditorComponent from '../components/paint-editor.jsx';
...
+    }
     componentDidMount () {
         document.addEventListener('keydown', this.props.onKeyPress);
     }
     componentWillUnmount () {
         document.removeEventListener('keydown', this.props.onKeyPress);
     }
+    handleUpdateSvg () {
+        if (!this.props.onUpdateSvg) {
+            return;
```
paulkaplan: You don't need to do this check since `props.onUpdateSvg` is a required prop.

PR #10, opened 2017-09-06
UI skeleton for the paint editor

fsih:
image

This PR pulls in buffered reader and CSS from scratch-gui and sets up the basic structure of the UI components for the paint editor. None of the buttons have any functionality yet.

(The pulled in components should eventually end up in a shared library, but we've decided to put off building that until the paint editor is more complete)

### DIFF
--- a/src/components/forms/buffered-input-hoc.jsx
+++ b/src/components/forms/buffered-input-hoc.jsx
@@ -1,3 +1,7 @@
+/* DO NOT EDIT
+@todo This file is copied from GUI and should be pulled out into a shared library.
+See https://github.com/LLK/scratch-paint/issues/13 */
+
 import bindAll from 'lodash.bindall';
 import PropTypes from 'prop-types';
 import React from 'react';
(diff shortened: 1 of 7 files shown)


src/containers/paper-canvas.jsx
```diff
@@ -60,7 +62,11 @@ class PaperCanvas extends React.Component {
     render () {
         return (
             <canvas
+                className={styles.paperCanvas}
+                height="400px"
                 ref={this.setCanvas}
+                resize="false"
```
paulkaplan: I get a react warning about this ,`Unknown prop `resize` on <canvas> tag.`.

src/components/paint-editor.css
```diff
@@ -0,0 +1,136 @@
...
+}
+
+.button > img {
+    flex-grow: 1;
+    max-width: 100%;
+    max-height: 100%;
+    min-width: 1.5rem;
+}
+
+.trim-button {
```
paulkaplan: Can you take out from here any of the sound-editor specific things that aren't being used?

PR #11, opened 2017-09-07
Add fill and stroke color state

fsih:
This PR should go in after https://github.com/LLK/scratch-paint/pull/10

For diff, see: https://github.com/fsih/scratch-paint/compare/structure...fsih:fillStrokeColor?expand=1

This PR adds components and state for the fill and stroke color, and pipes the state through to the existing tools. The components are just placeholder text inputs that take hex codes.

### DIFF
--- a/src/components/fill-color-indicator.jsx
+++ b/src/components/fill-color-indicator.jsx
@@ -20,7 +20,6 @@ const FillColorIndicatorComponent = props => (
     <div className={styles.inputGroup}>
         <Label text={props.intl.formatMessage(messages.fill)}>
             <BufferedInput
-                tabIndex="1"
                 type="text"
                 value={props.fillColor}
                 onSubmit={props.onChangeFillColor}
--- a/src/components/paint-editor.jsx
+++ b/src/components/paint-editor.jsx
@@ -44,7 +44,6 @@ class PaintEditorComponent extends React.Component {
                     <div className={styles.inputGroup}>
                         <Label text={this.props.intl.formatMessage(messages.costume)}>
                             <BufferedInput
-                                tabIndex="1"
                                 type="text"
                                 value="meow"
                             />
--- a/src/components/stroke-color-indicator.jsx
+++ b/src/components/stroke-color-indicator.jsx
@@ -20,7 +20,6 @@ const StrokeColorIndicatorComponent = props => (
     <div className={styles.inputGroup}>
         <Label text={props.intl.formatMessage(messages.stroke)}>
             <BufferedInput
-                tabIndex="1"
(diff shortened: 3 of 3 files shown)


src/containers/fill-color-indicator.jsx
```diff
@@ -0,0 +1,31 @@
...
+import {connect} from 'react-redux';
+import {changeFillColor} from '../reducers/fill-color';
+import FillColorIndicatorComponent from '../components/fill-color-indicator.jsx';
+
+const FillColorIndicator = props => (
+    <FillColorIndicatorComponent
+        fillColor={props.fillColor}
+        onChangeFillColor={props.handleChangeFillColor}
+    />
+);
```
paulkaplan: When creating a container for a component like this, you actually do not have to create a new component first. e.g. remove lines 7-17 and then replace the rest with this

```javascript
 const mapStateToProps = state => ({
     fillColor: state.scratchPaint.fillColor
 });

 const mapDispatchToProps = dispatch => ({
     onChangeFillColor: fillColor => {
         dispatch(changeFillColor(fillColor));
     }
 });
 
 export default connect(
     mapStateToProps,
     mapDispatchToProps
 )(FillColorIndicatorComponent);
```

Note the change to the naming in `mapDispatchToProps`, and how the end uses `connect` directly with the component.

src/reducers/fill-color.js
```diff
@@ -0,0 +1,33 @@
+import log from '../log/log';
+
+const CHANGE_FILL_COLOR = 'scratch-paint/fill-color/CHANGE_FILL_COLOR';
+const initialState = '#000';
+// Matches hex colors
+const regExp = /^#([0-9a-f]{3}){1,2}$/i;
```
paulkaplan, line 6: regexp doesn't support capital letters?

src/containers/stroke-color-indicator.jsx
```diff
@@ -0,0 +1,31 @@
...
+import {connect} from 'react-redux';
+import {changeStrokeColor} from '../reducers/stroke-color';
+import StrokeColorIndicatorComponent from '../components/stroke-color-indicator.jsx';
+
+const StrokeColorIndicator = props => (
+    <StrokeColorIndicatorComponent
+        strokeColor={props.strokeColor}
+        onChangeStrokeColor={props.handleChangeStrokeColor}
+    />
+);
```
paulkaplan: Same comment as with the fill case.

src/reducers/fill-color.js
fsih: Checked that capital letters are working

PR #14, opened 2017-09-14
Select

fsih:
Resolves https://github.com/LLK/scratch-paint/issues/18

This pulls in the select tool from papergrapher. It also pulls in the many helper files that the select tool uses, which are in a new folder helper/

The select tool is refactored into 5 tools:
- SelectMode, the React component, handles the selection box that is drawn when you click in an empty area and drag. It also sets the highlighted "hoverItem" as the mouse moves. In onMouseDown, if you hit something instead of an empty space, we switch to the bounding box tool
    - BoundingBoxTool handles drawing the bounding box around selected items. In onMouseDown, depending on what part of the bounding box is hit, it switches to the RotateTool, ScaleTool, or MoveTool.
        - MoveTool translates the item
        - ScaleTool resizes the item
        - RotateTool rotates the item about its center
        These tools also have special properties if alt or shift are held.

Refactoring out the BoundingBoxTool is designed to allow the shape modes to reuse it without having to enter select mode.

### DIFF
--- a/src/containers/select-mode.jsx
+++ b/src/containers/select-mode.jsx
@@ -9,7 +9,6 @@ import {setHoveredItem, clearHoveredItem} from '../reducers/hover';
 
 import SelectTool from '../helper/selection-tools/select-tool';
 import SelectModeComponent from '../components/select-mode.jsx';
-import paper from 'paper';
 
 class SelectMode extends React.Component {
     constructor (props) {
@@ -25,8 +24,8 @@ class SelectMode extends React.Component {
         }
     }
     componentWillReceiveProps (nextProps) {
-        if (this.tool && nextProps.hoveredItem !== this.props.hoveredItem) {
-            this.tool.setPrevHoveredItem(nextProps.hoveredItem);
+        if (this.tool && nextProps.hoveredItemId !== this.props.hoveredItemId) {
+            this.tool.setPrevHoveredItemId(nextProps.hoveredItemId);
         }
 
         if (nextProps.isSelectModeActive && !this.props.isSelectModeActive) {
@@ -57,7 +56,7 @@ class SelectMode extends React.Component {
 SelectMode.propTypes = {
     clearHoveredItem: PropTypes.func.isRequired,
     handleMouseDown: PropTypes.func.isRequired,
-    hoveredItem: PropTypes.instanceOf(paper.Item),
+    hoveredItemId: PropTypes.number,
     isSelectModeActive: PropTypes.bool.isRequired,
     onUpdateSvg: PropTypes.func.isRequired,
     setHoveredItem: PropTypes.func.isRequired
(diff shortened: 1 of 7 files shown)


src/containers/paper-canvas.css
```diff
@@ -4,4 +4,5 @@
     margin: auto;  
     position: relative;
     background-color: #fff;
+	image-rendering: pixelated;
```
rschamp: What does this do? It would be good to include a comment here to explain why this is here.

src/containers/select-mode.jsx
```diff
@@ -0,0 +1,169 @@
...
+            'deactivateTool',
+            'getHitOptions'
+        ]);
+        this._hitOptions = {
+            segments: true,
+            stroke: true,
+            curves: true,
+            fill: true,
+            guide: false
+        };
```
rschamp: A comment to explain the `_hitOptions` object and/or the options would be helpful here.

src/containers/select-mode.jsx
```diff
@@ -0,0 +1,169 @@
...
+        if (nextProps.isSelectModeActive && !this.props.isSelectModeActive) {
+            this.activateTool();
+        } else if (!nextProps.isSelectModeActive && this.props.isSelectModeActive) {
+            this.deactivateTool();
+        }
+    }
+    shouldComponentUpdate () {
+        return false; // Static component, for now
+    }
+    getHitOptions (preselectedOnly) {
```
rschamp: A comment to explain this method would be helpful as well, especially since it can modify what it's returning. That behavior looks like it could cause bugs, since it's vulnerable to race conditions.

src/containers/select-mode.jsx
```diff
@@ -0,0 +1,169 @@
...
+    }
+    activateTool () {
+        selectRootItem();
+        this.boundingBoxTool.setSelectionBounds();
+        this.tool = new paper.Tool();
+
+        // Define these to sate linter
+        const selectMode = this;
+
+        this.tool.onMouseDown = function (event) {
```
rschamp: Is there a reason for the `selectMode` variable other than to carry in `this`? If this were defined as an arrow function, you could just use `this` within it.

src/containers/selection-hov.jsx
```diff
@@ -0,0 +1,46 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import {connect} from 'react-redux';
+import paper from 'paper';
+
+const SelectionHOV = function (WrappedComponent) {
```
rschamp: What's `HOV`?

src/containers/select-mode.jsx
fsih: Moved variable creation into the function.

src/containers/selection-hov.jsx
fsih: High occupancy vehicle

PR #45, opened 2017-10-03
Selection updates color indicators, and color indicators update selection

fsih:
Depends on https://github.com/LLK/scratch-paint/pull/15
Fixes https://github.com/LLK/scratch-paint/issues/20
Results of this PR:
1. When some items are selected, the fill color, stroke color, and stroke width indicators update to show the values of the selection. It might show "mixed".
2. When some items were selected and you change the fill color, stroke color, or stroke width, the selected items change to have that value.

Since we now need the UI to react after a change in selection, the selection had to be stored in the Redux state in some way. I decided to do this by storing paper.project.selectedItems in the state at state.paintEditor.selectedItems.
- This is sufficient for now, but might not be if we want to track changes in individual selected points. Will cross that bridge if we come to it.
- Annoyingly you have to remember to dispatch the state update action every time you select anything in Paper. The clearSelection function helps you remember by taking the dispatch function as a parameter, but anywhere else the selection is changed you have to call it yourself.
- Since the dispatcher belongs to React components, you have to pass it down into the helper functions everywhere. Is there a way to give the helper functions their own dispatcher?

### DIFF
--- a/src/helper/style-path.js
+++ b/src/helper/style-path.js
@@ -9,7 +9,7 @@ const MIXED = 'scratch-paint/style-path/mixed';
  * @param {string} colorString New color, css format
  */
 const applyFillColorToSelection = function (colorString) {
-    const items = getSelectedLeafItems;
+    const items = getSelectedLeafItems();
     for (const item of items) {
         if (isPGTextItem(item)) {
             for (const child of item.children) {


src/containers/blob/blob.js
```diff
@@ -45,7 +52,18 @@ class Blobbiness {
...
+        const oldStrokeColor = this.options ? this.options.strokeColor : null;
+        const oldStrokeWidth = this.options ? this.options.strokeWidth : null;
+        // If values are mixed, it means the color was set by a selection contained multiple values.
+        // In this case keep drawing with the previous values if any. (For stroke width, null indicates)
+        // mixed, because stroke width is required to be a number)
+        this.options = {
+            ...options,
+            fillColor: options.fillColor === MIXED ? oldFillColor : options.fillColor,
+            strokeColor: options.strokeColor === MIXED ? oldStrokeColor : options.strokeColor,
+            strokeWidth: options.strokeWidth === null ? oldStrokeWidth : options.strokeWidth
```
paulkaplan, line 65: So mixed colors are represented by `MIXED` but mixed stroke widths are `null`? Should they be represented the same way?

src/containers/blob/blob.js
```diff
@@ -45,7 +52,18 @@ class Blobbiness {
      * @param {?number} options.strokeWidth Width of the brush outline.
      */
     setOptions (options) {
-        this.options = options;
+        const oldFillColor = this.options ? this.options.fillColor : 'black';
+        const oldStrokeColor = this.options ? this.options.strokeColor : null;
+        const oldStrokeWidth = this.options ? this.options.strokeWidth : null;
+        // If values are mixed, it means the color was set by a selection contained multiple values.
+        // In this case keep drawing with the previous values if any. (For stroke width, null indicates)
```
paulkaplan: Looks like a typo `)` at the end of this line

src/containers/line-mode.jsx
```diff
@@ -287,6 +295,12 @@ const mapDispatchToProps = dispatch => ({
     changeStrokeWidth: strokeWidth => {
         dispatch(changeStrokeWidth(strokeWidth));
     },
+    clearSelectedItems: () => {
+        dispatch(clearSelectedItems());
+    },
+    setSelectedItems: () => {
+        dispatch(setSelectedItems(getSelectedItems(true /* recursive */)));
```
paulkaplan: It might be better to just have `getSelectedItems` take an object like `{recursive: true}` instead of always using a comment.

src/helper/style-path.js
```diff
@@ -0,0 +1,193 @@
...
+                            selectionStrokeWidth = null;
+                        }
+                    }
+                }
+            }
+        } else if (!isGroup(item)) {
+            if (item.fillColor) {
+                // hack bc text items with null fill can't be detected by fill-hitTest anymore
+                if (isPointTextItem(item) && item.fillColor.toCSS() === 'rgba(0,0,0,0)') {
+                    itemFillColorString = null;
```
paulkaplan, line 134: is `null` a valid color value? Aren't the only valid values a color string and `MIXED`?

Also, what does the comment mean?

src/helper/style-path.js
```diff
@@ -0,0 +1,193 @@
...
+                } else {
+                    itemFillColorString = item.fillColor.toCSS();
+                }
+            }
+            if (item.strokeColor) {
+                itemStrokeColorString = item.strokeColor.toCSS();
```
paulkaplan, line 140: Should we store colors as separate rgba components or something more standardized? It'll be hard to get strings to work with the component color picker, unless they are all hex strings or something similarly rigid.

src/containers/blob/blob.js
fsih: Thanks!

src/containers/blob/blob.js
fsih: I had to do this because the stroke width indicator component requires a proptype of number, and I couldn't pass in the string "MIXED".  I agree it's pretty confusing. I could define another const MIXED_STROKE_WIDTH that's a number?

src/helper/style-path.js
fsih: Yes, null means transparent.

I have no clue what the comment means. I'll probably have to figure it out when I implement text.

src/containers/line-mode.jsx
fsih: I add the comment everywhere because it's unclear what the argument means, but it's actually still pretty unclear even with the comment. The "recursive" argument actually substantially changes what the function does, so I decided it would be clearer to break the function into 2 functions, one that's equivalent to recursive true and one for recursive false.

src/helper/style-path.js
fsih: Yes, we should do this! Filed issue https://github.com/LLK/scratch-paint/issues/53

src/containers/blob/blob.js
paulkaplan: sure, in fact making both constants named more explicitly would be good. Another solution would be to use `proptypes.oneOf([number, string])` if you want to allow both types.

Issue #47, opened 2017-10-05
Top row button disabled states


fsih:

fsih:
group/ungroup is in helper/group.js. Front/back is in helper/order.js

mewtaylor:
Implemented in #104

PR #48, opened 2017-10-05
Add undo/redo

fsih:
This PR adds undo/redo states, fixing https://github.com/LLK/scratch-paint/issues/21

This PR depends on https://github.com/LLK/scratch-paint/pull/45
See diff at: https://github.com/fsih/scratch-paint/compare/colorChanger...fsih:addUndo?expand=1

Undo/Redo buttons are hooked up. Undo snapshotting is added in onUpdateSvg. onUpdateSvg updates the stage to match what's in the paint editor, so it should be the right granularity for undo snapshots, which also want to change when anything visible changes. I also piped onUpdateSvg through everywhere and made it more careful about checking for a change before snapshotting.

### DIFF
--- a/src/helper/selection.js
+++ b/src/helper/selection.js
@@ -171,7 +171,7 @@ const _deleteItemSelection = function (items, onUpdateSvg) {
     }
     
     // @todo: Update toolbar state on change
-    if (items.lenth > 0) {
+    if (items.length > 0) {
         paper.project.view.update();
         onUpdateSvg();
     }


src/helper/undo.js
```diff
@@ -0,0 +1,46 @@
...
+// modifed from https://github.com/memononen/stylii
+import paper from 'paper';
+
+const performSnapshot = function (dispatchPerformSnapshot) {
+    dispatchPerformSnapshot({
+        json: paper.project.exportJSON({asString: false})
+    });
+
+    // @todo enable/disable buttons
+    // updateButtonVisibility();
```
paulkaplan, line 11: I don't think you should need to manually update the undo/redo button state, you should be able to compute a `canUndo`/`canRedo` property from the store in the render method for the buttons.

fsih: I'm planning to have button enabled/disabled state in a separate PR eventually and the logic will move to the components. It's filed here: https://github.com/LLK/scratch-paint/issues/47

src/components/paint-editor.jsx
```diff
@@ -74,12 +76,12 @@ class PaintEditorComponent extends React.Component {
                         <button
                             className={styles.button}
                         >
-                            Front
+                            <font color="silver"> Front </font>
```
rschamp: 🙀  Why not in the CSS?

fsih: I put in a couple of quick fixes to make the interface look cleaner for the oct 5 "playtest". I will take them back out.

PR #56, opened 2017-10-12
Switch paper dependencies to our own fork

fsih:
Switch to a fork of paper.js that's customized to have our own selection style.

Ideally we will only depend on this fork temporarily. We should be able to use paper.js again after https://github.com/paperjs/paper.js/pull/1380 , which might take awhile


fsih:
Closing because now that LLK/scratch-paint is public, my fork got unattached

Issue #31, opened 2017-09-26
Color picker dropdown


fsih:
This is the panel that appears when the color picker component is clicked, with HSV sliders

PR #59, opened 2017-10-13
Color picker

paulkaplan:
For dependency reasons, this depends on https://github.com/LLK/scratch-paint/pull/58, You can ignore the first two commits.

Issue https://github.com/LLK/scratch-paint/issues/31
Early version of the slider color picker first seen in scratch-blocks. 

paint-color-picker

There is a fair amount that is still rough about the components, fixed sizing and stuff. There is also no transparent "strike-through" icon for the transparent swatches.

But it is a start!


paulkaplan:
@fsih sorry i made a new PR because of the react 16 dependency, but I made all the changes from your comments on the old PR e2e4823

fsih (approved):
Cool, I guess I will ask later about px vs rem in the other css

PR #70, opened 2017-10-18
Line tool snapping

fsih:
When lines snap together, they were using selection state to indicate snapping. However, selection state now determines the state of the color pickers, and we don't want colors to start changing around based on snapping suggestions. So I've switched line tool snapping to draw a guide instead of using selection.

In addition, I've moved shared functions between line and pen tool into helper functions in preparation for adding the pen tool.

snapping

### DIFF
--- a/src/containers/line-mode.jsx
+++ b/src/containers/line-mode.jsx
@@ -4,11 +4,12 @@ import React from 'react';
 import {connect} from 'react-redux';
 import bindAll from 'lodash.bindall';
 import Modes from '../modes/modes';
-import {clearSelection, getSelectedLeafItems} from '../helper/selection';
+import {clearSelection} from '../helper/selection';
+import {endPointHit, touching} from '../helper/snapping';
+import {drawHitPoint, removeHitPoint} from '../helper/guides';
 import {MIXED} from '../helper/style-path';
 import {changeMode} from '../reducers/modes';
-import {changeStrokeWidth} from '../reducers/stroke-width';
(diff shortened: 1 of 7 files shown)


paulkaplan (approved):
These changes look good.  

I had some trouble reproducing some of the small segment edge cases without zooming. 

Additionally, I think some things like snap tolerance will have to be scaled with zooming instead of having absolute tolerances. 

And finally I think we should keep an eye out in workshops for the desire to snap to a non-terminal path point. I know that kind of thing kind of "breaks" the fill/stroke model, but I suspect it might be a common and somewhat frustrating limitation.

paulkaplan:
Oh one other thing. It looks like the line tool isn't clearing the curve handles when it you start a line at the end of a curved segment. For example, create a curve with the pen, erase through it and draw a line at one of the start points. I'm not sure if this is a problem with the line tool or the eraser though.

fsih:
- Small segment connections fixed
- Tolerances are usually scaled by zoom
- I was looking at the 2.0 editor and I don't think it lets you snap to corners that aren't endpoints. Snapping seems to be buggy; sometimes it shows you the little snapping circle, but when you click it doesn't actually join the lines. So it might seem like you can snap to corners, but it's because the lines forming the corner were 2 separate lines

fsih:
Fixed your second comment. Thanks for finding these!

Since I added a couple commits, you might want to review again

paulkaplan (approved):

PR #71, opened 2017-10-18
Pen tool

fsih:
Depends on https://github.com/LLK/scratch-paint/pull/70/files
See diff at https://github.com/fsih/scratch-paint-1/compare/lineToolSnapping...penTool?expand=1

Add the pen tool

pen

### DIFF
--- a/src/containers/pen-mode.jsx
+++ b/src/containers/pen-mode.jsx
@@ -5,7 +5,6 @@ import bindAll from 'lodash.bindall';
 import Modes from '../modes/modes';
 
 import {changeMode} from '../reducers/modes';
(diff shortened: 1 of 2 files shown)


paulkaplan (approved):
Looks good! And the smoothing and connection stuff seems to work very well. 

Some things to watch out for in workshops: How does the smoothing interact with zooming? It isn't being scaled according to zoom level, not sure if that might get frustrating.

paulkaplan:
There was that other issue with drawing a line starting from a pen segment that I mentioned on the line PR, but that probably isn't related to this (either line or eraser probably).

PR #72, opened 2017-10-19
Adding scratch-l10n dependency

chrisgarrity:
* Added .tx configuation to be able to send source strings to transifex
* added dependency on scratch-l10n 2.x, and removed everything related to generating the l10n messages locally
* updated playground intl reducer to use paint messages from l10n package


fsih (approved):
Thanks!

How do I test with languages other than English?

PR #78, opened 2017-10-19
Add props for name, onUpdateName, and svgId

rschamp:
Adds new props for the PaintEditor component so that the costume name can be updated. Update the playground to account for these.

With this change, also only reset undo state when `svgId` changes, not the `svg` prop.

### DIFF
--- a/src/containers/paint-editor.jsx
+++ b/src/containers/paint-editor.jsx
@@ -54,11 +54,13 @@ class PaintEditor extends React.Component {
     render () {
         return (
             <PaintEditorComponent
+                name={this.props.name}
                 rotationCenterX={this.props.rotationCenterX}
                 rotationCenterY={this.props.rotationCenterY}
                 svg={this.props.svg}
                 onRedo={this.handleRedo}
                 onUndo={this.handleUndo}
+                onUpdateName={this.props.onUpdateName}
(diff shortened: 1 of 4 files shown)


src/containers/paint-editor.jsx
```diff
@@ -54,21 +54,25 @@ class PaintEditor extends React.Component {
     render () {
         return (
             <PaintEditorComponent
+                name={this.props.name}
                 rotationCenterX={this.props.rotationCenterX}
                 rotationCenterY={this.props.rotationCenterY}
                 svg={this.props.svg}
```
fsih, line 60: Whenever the paint editor gets new props, if it gets an SVG it would reload and clear the undo state as if you switched to a different costume.  I think we have to design a way to avoid this if you have to pass in name.

rschamp: Fixed in https://github.com/LLK/scratch-paint/pull/78/commits/138718008c7e78c3da3e15919f98e45825de544d

fsih (approved):
I think the paint editor in GUI won't update properly when we put this in, so maybe we should wait to put this and the GUI update in around the same time.

rschamp:
@fsih Yes, I think we just have to make sure to merge the GUI PR immediately after this one is published.  So I'm going to do that now.

Issue #40, opened 2017-09-26
Paint editor UI


fsih:
- Bring in icons
- Proper size of items and spacing between items

fsih:
Here is a super vague 'everything UI' bug; feel free to break it out into more reasonable pieces

src/components/paint-editor.jsx is where the outline of the UI lives now. Let me know if I can help!

mewtaylor:
@fsih this seems done too, at this point perhaps?

Issue #13, opened 2017-09-11
Shared library for components in paint editor and GUI


fsih:
Much of the UI, such as buttons and input fields, should be identical in the paint editor and scratch-gui. Those pieces should be componentized and moved into a shared repo.

PR #79, opened 2017-10-19
Add svg icons to tool select components

mewtaylor:
This also adds the following new things to the repo:
* a generic `tool-select` component which can be composed by each of the individual tools for presentation purposes.
* a new dev dependency on `svg-url-loader` for adding svgs to components
* the button component, copied over from `scratch-gui`
* moves components into directories so that they can be stored alongside svgs, and css files if applicable

Towards #40.


src/components/button/button.jsx
```diff
@@ -0,0 +1,33 @@
+// Copied from scratch-gui for now
```
fsih: Let's keep these comments the same so we can find all these files easily later

Right now in some of the other files the comment is:
/* DO NOT EDIT
@todo This file is copied from GUI and should be pulled out into a shared library.
See https://github.com/LLK/scratch-paint/issues/13 */

src/components/button/button.jsx
```diff
@@ -0,0 +1,33 @@
...
+    ...props
+}) => (
+    <span
+        className={classNames(
+            styles.button,
+            className
+        )}
+        role="button"
+        onClick={onClick}
+        {...props}
```
fsih, line 24: It looks like {...props} would be empty?

src/components/oval-mode/oval.svg
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg width="20px" height="20px" viewBox="0 0 20 20" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 43.2 (39069) - http://www.bohemiancoding.com/sketch -->
+    <title>oval</title>
+    <desc>Created with Sketch.</desc>
```
fsih, line 5: These SVG descs aren't surfaced somehow in the UI are they? It would be weird if you were using a screen reader and it kept saying "Created with Sketch"

src/components/button/button.jsx
mewtaylor: awesome, thanks!

src/components/button/button.jsx
mewtaylor: It is for current instances – the idea was to pass props to the button itself that are custom for a given implementation, should it arise.

I can take this out, though, if we don't want it for now!

fsih: Oh cool, I didn't know that was a thing you could do

fsih (dismissed):

src/components/oval-mode/oval.svg
paulkaplan: I don't think so, because I think they end up being encoded as data uri images.

fsih (approved):

PR #86, opened 2017-10-20
Delete shapes

fsih:
Depends on https://github.com/LLK/scratch-paint/pull/85
Diff: https://github.com/fsih/scratch-paint-1/compare/ovalTool...fsih:deleteShapes?expand=1

Hook up delete key to work when things are selected in shape tools

### DIFF
--- a/src/components/paint-editor/paint-editor.jsx
+++ b/src/components/paint-editor/paint-editor.jsx
@@ -11,7 +11,6 @@ import SelectMode from '../../containers/select-mode.jsx';
 import LineMode from '../../containers/line-mode.jsx';
 import PenMode from '../../containers/pen-mode.jsx';
 import RectMode from '../../containers/rect-mode.jsx';
-import RoundedRectMode from '../../containers/rounded-rect-mode.jsx';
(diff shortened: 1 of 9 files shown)


paulkaplan (approved):
Not going to lie, this makes me kind of scared for the future of the selection tool vs. being selected. Are there other things that we'll have to move over for this "not quite selected" mode?

That being said, I understand this is the kind of thing it is best to get working and get people in front of it. So I think this is ok for now, but future us beware – this is a slippery slope.

fsih:
I don't think it's that ill defined. Selection isn't really a mode so much as a bunch of totally separate modes you can choose based on where you mouse down. Illustrator has this kind of behavior, where the newly drawn shape is selected without changing to a selection tool (not saying Illustrator UI makes any sense, but it's a reference point). Selecting newly drawn shapes was a feature that we felt we really needed after playtesting, and I think it feels more natural than not having it.

Delete should maybe be moved higher up eventually? Whatever tool you're in, you probably expect delete to delete the selection. We still have to deal with how delete should work when your cursor is in a text field.

PR #87, opened 2017-10-20
Fix color matching logic in apply stroke color

fsih:
It had an issue with the logic where if the color string was null and the existing stroke color was black, it would think the colors matched, so you couldn't switch between black and transparent.


mewtaylor (approved):
Looks good to me!

Issue #29, opened 2017-09-26
Build new UI components


fsih:
Text-icon button in paint editor has elements of text button and icon button which are present in GUI. These should be consolidated.

fsih:
Currently, all the files that exactly match files in GUI have this header:
/* DO NOT EDIT
@todo This file is copied from GUI and should be pulled out into a shared library.
See https://github.com/LLK/scratch-paint/issues/13 */

mewtaylor:
@fsih should we mark this as closed now you think?

fsih:
Yep!

mewtaylor:
Closed in #88

PR #88, opened 2017-10-23
Add UI formatting for edit fields

mewtaylor:
This adds formatting and icons to the edit field buttons at the top of the ui, as well as some additional components for abstraction – such as `InputGroup`, `ButtonGroup` and `EditFieldButton` (which is a subcomponent of the paint editor).

Towards #29 and #40


src/components/paint-editor/edit-field-button/edit-field-button.css
```diff
@@ -0,0 +1,31 @@
+@import "../../../css/colors.css";
+@import "../../../css/units.css";
+
+$border-radius: 0.25rem;
+
+.mod-edit-field {
```
paulkaplan, line 6: What does the `.mod-` class prefix indicate? Haven't seen it here before.

paulkaplan (dismissed):
Changes LG, I just had a small question.

Looks like things like font sizing aren't to spec yet, but I assume that'll come later? If yes, good to merge.

src/components/paint-editor/paint-editor.jsx
```diff
@@ -47,82 +62,123 @@ class PaintEditorComponent extends React.Component {
...
+                            />
+                            <EditFieldButton
+                                imgAlt="Send to Back Icon"
+                                imgSrc={sendBackIcon}
+                                title="Back"
+                                onClick={function () {}}
+                            />
+                        </InputGroup>
+
+                        {/* To be rotation point */}
```
fsih, line 152: This one is not in the plan for the playtest, so can we hide it for now?

src/components/paint-editor/edit-field-button/edit-field-button.css
mewtaylor: It's from the trello style guide. If you have a general style for something – in this case, a button – and are applying style that is more specific to a particular use case – here, the edit fields – then the convention is to prefix the class name for that use case with `mod-` to show that it's a modification of the base class.

I realize now that I just kind of did that instinctually since that's how www is set up, but could be off base. Is that not the best way to name this kind of class?

fsih (approved):

PR #89, opened 2017-10-23
Add group/ungroup

fsih:
Hook up group and ungroup buttons.

The bounding box tool needs to know when group/ungroup happen, so that it can redraw the selection bounds. It now listens to scratchPaint.selectedItems in the redux state and redraws the selection box when the selection changes. I had to pipe this state through multiple layers since bounding-box-tool is kind of far away from the components layer. I also added updates to the selection state in more places to keep it more accurate, for instance after reloading an old state via undo.

### DIFF
--- a/src/components/paint-editor/paint-editor.jsx
+++ b/src/components/paint-editor/paint-editor.jsx
@@ -150,14 +150,14 @@ class PaintEditorComponent extends React.Component {
                         </InputGroup>
 
                         {/* To be rotation point */}
-                        <InputGroup>
+                        {/* <InputGroup>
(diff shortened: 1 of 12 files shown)


paulkaplan:
Did you want to handle the enable/disable states of the group/ungroup buttons based on selection in a different PR? Is that waiting on something else?

fsih:
Enabled/disabled states aren't waiting on anything. I haven't implemented them for any of the buttons, so I was going to do them together in their own PR.

paulkaplan (dismissed):
Looks good!

PR #91, opened 2017-10-23
Layer order

fsih:
Implement the forward/backward/front/back buttons

### DIFF
--- a/src/helper/order.js
+++ b/src/helper/order.js
@@ -1,0 +1,70 @@
+import {getSelectedRootItems} from './selection';
+
+const bringToFront = function (onUpdateSvg) {
+    const items = getSelectedRootItems();
+    for (const item of items) {
+        item.bringToFront();
+    }
+    onUpdateSvg();
+};
+
+const sendToBack = function (onUpdateSvg) {
+    const items = getSelectedRootItems();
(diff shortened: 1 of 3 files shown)


src/helper/order.js
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    const previousSibling = items[0].previousSibling;
+    for (const item of items) {
+        item.insertBelow(previousSibling);
+    }
+    onUpdateSvg();
+};
+
+const shouldShowSendBackward = function () {
```
paulkaplan, line 47: In the future, I think it is good practice to only submit code that is actually used in each PR. I know this code is very related, and will be used in follow-up PRs, but it complicates the git history to have features spread between different PRs. For example, it would make more sense to submit this code when the UI parts of enable/disable are also used. It leaves less unreachable code around in the git history.

paulkaplan (approved):
Looks good! 

Had just one style tip but doesn't require any changes.

src/helper/order.js
fsih: You're right, thanks. I have been being messy about bringing over the helper functions and then have to prune them later. I will do that in the future.

Issue #94, opened 2017-10-24
Change default mode/tool order


paulkaplan:
Default tool order update, from @carljbowman
<img width="44" alt="screen shot 2017-10-24 at 3 47 13 pm" src="https://user-images.githubusercontent.com/654102/31964678-cec8612e-b8d2-11e7-86d9-0a17d03b9430.png">

PR #95, opened 2017-10-24
Update default mode to select and mode button order

paulkaplan:
Fixes https://github.com/LLK/scratch-paint/issues/94


fsih (approved):
Thanks!

PR #97, opened 2017-10-24
Add the background guide layer

fsih:
image

- Add another guide layer behind the paint layer with the grid and center-marking crosshair
- Change instances of paper.project.getItems to our own getItems that doesn't include guide layers.


towerofnix:
>Need to find a way to make the grid of 6000 tiny squares more performant.

Maybe avoid using `paper` here - it looks like a plain old JavaScript canvas is more appropriate. It might be most appropriate to make (either with JS canvas or as an asset) a "2x2"-square image which can be tiled and set that as a background image (either a plain old CSS background image, or just have a single `paper` Rectangle object whose background is this tile image, repeating, though I don't know if that's simple to implement?).

Edit: And at the very least, definitely make sure you're caching it! :) (I don't know if you are or not yet; I'm not familiar with `scratch-paint` or `paper`.)

fsih:
Fixed the performance issue by using a raster instead. Thanks for the suggestions!

(We want the background to live in paper so that we can scale it when we add zoom)

towerofnix:
>We want the background to live in paper so that we can scale it when we add zoom.

I figured, yeah. Glad to help!

paulkaplan (approved):
LG!

@carljbowman this background seem a bit heavy, like pretty dark gray and very fine grained. But that's just me...

fsih:
@carljbowman I agree, the squares look really tiny to me and kind of painful to stare at, but it does look pretty much the same in Scratch 2.0

Issue #36, opened 2017-09-26
Brush size picker


fsih:

fsih:
The state for brush and eraser modes is already in; see reducers/brush-mode and reducers/eraser-mode. You would just want to make an input that sets that state and take out the onScroll listeners in those components that currently control that state.

mewtaylor:
Implemented in #99

PR #99, opened 2017-10-25
Implement GH-36 and GH-37: Add basic eraser and brush size picker

mewtaylor:
this implements the first version of #36 and #37, in which the brush size picker is a numeric input field rather than a slider.

Other things that this PR does:
* Creates a new `ModeToolsComponent` in which the logic for conditionally showing mode tools is housed
* Adds a new variable to `units.css` for `grid-unit`, which is set to `.25rem`, the basic grid unit (done after discussion with @carljbowman)
* Changes the name of `EditSelectComponent` to `LabeledIconButton` after discussion with @paulkaplan 
* reins in some formatting for the tools after a css pass with @carljbowman 
* Adjusts `BufferedInputHOC` to change the value `onChange()`, not just `onBlur()`

**Note:** We will still need to implement the dropdown slider for eraser and brush too
**Another Note:** This PR also includes icons for not-yet-implemented mode tools, such as for reshape. My plan was to add these, and then disable them when I submit a PR to handle disabled states, but am ok with removing them if that's preferable

### DIFF
--- a/src/components/forms/buffered-input-hoc.jsx
+++ b/src/components/forms/buffered-input-hoc.jsx
@@ -2,6 +2,10 @@
 @todo This file is copied from GUI and should be pulled out into a shared library.
 See https://github.com/LLK/scratch-paint/issues/13 */
 
+/* ACTUALLY, THIS HAS BEEN EDITED ;)
+handleChange() was adjusted here to actually send a change to `onSubmit()` so that
+brush/eraser sizes change immediately on a numeric input*/
+
 import bindAll from 'lodash.bindall';
 import PropTypes from 'prop-types';
 import React from 'react';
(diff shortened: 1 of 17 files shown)


src/components/forms/buffered-input-hoc.jsx
```diff
@@ -39,6 +39,11 @@ export default function (Input) {
             this.setState({value: null});
         }
         handleChange (e) {
+            const isNumeric = typeof this.props.value === 'number';
```
paulkaplan, line 46: So... the top of this file says `DO NOT EDIT`...

Maybe add a comment about why it is being edited in the face of such clear instructions to the contrary ;)

src/components/mode-tools/mode-tools.jsx
```diff
@@ -0,0 +1,154 @@
...
+                    <img
+                        alt={eraserMessage}
+                        className={styles.modeToolsIcon}
+                        src={eraserIcon}
+                    />
+                </div>
+                <BufferedInput
+                    small
+                    max={MAX_STROKE_WIDTH}
+                    min="0"
```
fsih: Should be 1

src/components/mode-tools/mode-tools.jsx
```diff
@@ -0,0 +1,154 @@
...
+                    <img
+                        alt={brushMessage}
+                        className={styles.modeToolsIcon}
+                        src={brushIcon}
+                    />
+                </div>
+                <BufferedInput
+                    small
+                    max={MAX_STROKE_WIDTH}
+                    min="0"
```
fsih: Should be 1

fsih (changes requested):

src/components/forms/buffered-input-hoc.jsx
mewtaylor: lol my b thanks!

mewtaylor:
@fsih thanks, just updated them to be 1!

fsih (dismissed):

fsih (approved):

PR #104, opened 2017-10-25
Implement GH-47: Add disabled state for button component

mewtaylor:
This implements #47 by adding a disabled state to the button component, and then by applying it to top row buttons for group/ungroup, and front/back forward/backward.

### DIFF
--- a/src/components/paint-editor/paint-editor.jsx
+++ b/src/components/paint-editor/paint-editor.jsx
@@ -83,6 +83,7 @@ class PaintEditorComponent extends React.Component {
                                 <ButtonGroup>
                                     <Button
                                         className={styles.buttonGroupButton}
+                                        disabled={!this.props.canUndo()}
                                         onClick={this.props.onUndo}
                                     >
                                         <img
@@ -93,6 +94,7 @@ class PaintEditorComponent extends React.Component {
                                     </Button>
                                     <Button
                                         className={styles.buttonGroupButton}
+                                        disabled={!this.props.canRedo()}
                                         onClick={this.props.onRedo}
                                     >
                                         <img
@@ -245,6 +247,8 @@ class PaintEditorComponent extends React.Component {
 }
 
 PaintEditorComponent.propTypes = {
+    canRedo: PropTypes.func.isRequired,
+    canUndo: PropTypes.func.isRequired,
     intl: intlShape,
     name: PropTypes.string,
     onGroup: PropTypes.func.isRequired,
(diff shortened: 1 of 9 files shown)


paulkaplan:
For some reason things seem to initialize in a strange way. Until i drag something around, everything is disabled, even if something is selected.

odd-select-behavior

(this was recorded from within the GUI, fwiw)

src/components/paint-editor/paint-editor.jsx
```diff
@@ -62,126 +63,134 @@ class PaintEditorComponent extends React.Component {
...
-                                title="Ungroup"
-                                onClick={this.props.onUngroup}
-                            />
-                        </InputGroup>
+                            {/* Undo/Redo */}
+                            <InputGroup>
+                                <ButtonGroup>
+                                    <Button
+                                        className={styles.buttonGroupButton}
+                                        onClick={this.props.onUndo}
```
fsih: Not necessarily in this PR, but we should add disabled states for undo and redo too. (The functions aren't pre-existing in helper/undo, but if you look in reducers/undo, they would look like those functions that log "Can't undo" and "can't redo")

mewtaylor:
@paulkaplan ah that makes sense – since selecting is not causing a re-render, it's not getting a new check. Probably makes sense to have the call to `should...()` come from a reducer, and be dispatched on a select state perhaps? Thoughts @fsih ?

src/components/paint-editor/paint-editor.jsx
fsih: canUndo: return (state.pointer > 1)
canRedo: return (state.pointer > -1 && state.pointer !== state.stack.length - 1)

paulkaplan:
Another issue, I seem to get a bunch of these proptype warnings:
```
warning.js:33 Warning: Failed prop type: Invalid prop `disabled` of type `boolean` supplied to `ButtonComponent`, expected `string`.
```

fsih:
@mewtaylor That makes sense. Should show group/ungroup/front/back can listen on state.scratchPaint.selectedItems, which should update when the selection updates. I don't think it currently pushes a state change if you reorder items, so we will have to add that (calling setSelectedItems in bring to front/back etc.)

mewtaylor:
@paulkaplan I just pushed some changes to address the warning you were seeing, as well as the behavior issues you noticed too – thanks for that!

@fsih I'm happy to add enabled/disabled for undo/redo in this PR too, makes sense to me!

mewtaylor:
@paulkaplan @fsih I added the undo/redo disabled checks, lemme know what you think!

fsih (dismissed):
Looks great!

Just a small nit, when redo is enabled and undo is disabled, should the line between them be grayed out?
image

fsih (approved):

Issue #100, opened 2017-10-25
Hitting "backspace" on any input with a shape selected deletes it


paulkaplan:
Try to type in a new costume name, and then backspace a few times to fix a mistake. If you have an object selected on the canvas, it'll get deleted. 

One strategy would be to check the `event.target` and make sure that it isn't an `INPUT`.

PR #107, opened 2017-10-26
Don't delete when input field focused

fsih:
Fixes https://github.com/LLK/scratch-paint/issues/100
Moved delete up out of modes, so that it's always available and there's less duplicated code. Made the new delete function ignore key strokes when an input is focused.

### DIFF
--- a/src/containers/paper-canvas.jsx
+++ b/src/containers/paper-canvas.jsx
@@ -3,11 +3,14 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import {connect} from 'react-redux';
 import paper from '@scratch/paper';
+import Modes from '../modes/modes';
 
 import {performSnapshot} from '../helper/undo';
 import {undoSnapshot, clearUndoState} from '../reducers/undo';
 import {isGroup, ungroupItems} from '../helper/group';
 import {setupLayers} from '../helper/layer';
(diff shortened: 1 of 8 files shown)


src/containers/paper-canvas.jsx
```diff
@@ -45,6 +50,19 @@ class PaperCanvas extends React.Component {
...
         paper.remove();
+        document.removeEventListener('keydown', this.handleKeyDown);
+    }
+    handleKeyDown (event) {
+        if (event.target instanceof HTMLInputElement) {
+            // Ignore delete if a text input field is focused
+            return;
+        }
+        // Backspace, delete
+        if (event.key.toLowerCase() === 'delete' || event.key.toLowerCase() === 'backspace') {
```
paulkaplan: I don't think there is any capitalization variation with `event.key`, should just be `Delete` or `Backspace`

fsih: Done

paulkaplan (approved):
I have strong opinions about the need to deal with both backspace and delete, but i'll keep those to myself ;)

Looks good. I think it is good to move this out of individual components.

fsih:
Wait what's wrong with dealing with both backspace and delete? They're
separate buttons

Silly Mac user

PR #112, opened 2017-10-26
Fix several color state inconsistencies

paulkaplan:
1. Show transparent state as a red strikethrough. 
2. Add a box to make outline color button different from fill.
3. Add a "mixed color" icon.
4. Make the tools that require a fill, stroke or stroke width actually update those states so that they appear that way.
5. Make the tools that are incompatible with fill or stroke disable those inputs so that they are not confusing. 

We may want to revisit some or all of these choices after the workshop tomorrow.

### DIFF
--- a/src/components/color-button/color-button.css
+++ b/src/components/color-button/color-button.css
@@ -30,3 +30,8 @@
     color: #575e75;
     font-size: 0.75rem;
 }
+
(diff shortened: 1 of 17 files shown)


fsih (approved):
LG, thanks!

PR #116, opened 2017-10-27
Add basic zooming and panning

paulkaplan:
This PR adds some basic zoom in/out/reset buttons as well as basic panning functionality from the scroll wheel. Basically, zoom/pan is easy, but making scrollbars is hard. 

1. Zoom in, zoom out and reset buttons. Zoom by default tracks the selection, or zooms to the current view center.
2. Wheel zoom (holding ctrl/cmd while scrolling) which zooms to the point your mouse is over.
3. Panning, which for trackpads that do deltaX/Y just follows trackpad movement, for mousewheel it does up/down by default and left/right when you hold <kbd>Shift</kbd>. This particular key combination is actually because it is the default way the browser translates deltaX/deltaY when using a mouse. 

zoom-pan

My feeling with this is that it obviously isn't good enough to use with kids (because of the lack of scrollbars/discoverability) but it might be very helpful to have in order to make the internal workshop better by allowing people to make more things.

### DIFF
--- a/src/components/paint-editor/paint-editor.jsx
+++ b/src/components/paint-editor/paint-editor.jsx
@@ -100,7 +100,7 @@ class PaintEditorComponent extends React.Component {
                                         onClick={this.props.onUndo}
                                     >
                                         <img
-                                            alt="Undo Icon"
+                                            alt="Undo"
                                             className={styles.buttonGroupButtonIcon}
                                             src={undoIcon}
                                         />
@@ -118,7 +118,7 @@ class PaintEditorComponent extends React.Component {
                                         onClick={this.props.onRedo}
                                     >
                                         <img
-                                            alt="Redo Icon"
+                                            alt="Redo"
                                             className={styles.buttonGroupButtonIcon}
                                             src={redoIcon}
                                         />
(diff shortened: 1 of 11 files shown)


carljbowman:
YAAAAAAS! 

screen shot 2017-10-27 at 10 12 25 am

src/components/paint-editor/paint-editor.jsx
```diff
@@ -257,6 +256,41 @@ class PaintEditorComponent extends React.Component {
...
                         />
+                        {/* Zoom controls */}
+                        <InputGroup className={styles.zoomControls}>
+                            <ButtonGroup>
+                                <Button
+                                    className={styles.buttonGroupButton}
+                                    onClick={this.props.onZoomIn}
+                                >
+                                    <img
+                                        alt="Zoom In Icon"
```
fsih: Sorry I didn't catch this in earlier reviews. Our alt text shouldn't include 'icon'. I think the 'button' context is automatically added by screen readers so it can just say 'Zoom in'

src/containers/paper-canvas.jsx
```diff
@@ -114,13 +124,31 @@ class PaperCanvas extends React.Component {
             this.props.canvasRef(canvas);
         }
     }
+    handleWheel (event) {
```
fsih, line 127: Why is this called 'handleWheel'? Does it not work on touch pads and touch screens?

src/containers/paper-canvas.jsx
```diff
@@ -114,13 +124,31 @@ class PaperCanvas extends React.Component {
...
+        if (event.metaKey) {
+            // Zoom keeping mouse location fixed
```
fsih, line 129: It might be more useful to have this be the default behavior rather than the meta key behavior, since I don't think most people know you can convert your mouse scroll to a horizontal scroll, which means it'll be hard to move left/right otherwise. Of course we'd have to test it.

src/containers/paper-canvas.jsx
paulkaplan: Yup, the event is called `onWheel` so convention dictates the handler should be called `handleWheel`. The event is emitted by both mouse wheels and trackpads, but not on touch devices. We will need to handle those differently.

src/containers/paper-canvas.jsx
paulkaplan: Interesting idea. I think we'll have to see what it feels like with people. I think the trackpad behavior is good as the default, and I do not currently know a way to tell the difference between a real "mouse wheel" and a trackpad. If there was, I would totally swap them.

PR #118, opened 2017-10-27
Fix erasing with holes

fsih:
You can now make an eraser path that has a hole in it and it will still erase. Previously it did nothing.

There are still issues, especially if paths cross themselves.


fsih:
We felt like this was a pretty big issue worth working on before the internal playtest. Do we want to merge it for the external playtest, or hold off?

paulkaplan (approved):

PR #159, opened 2017-10-29
Only submit SVG changes when color picker closes

### LINKED ISSUE #125: Sprite renders twice when changing color in Vector Editor
**What happens:**
multi sprite rendering

**Repro**
1. Load / Create a sprite with multiple paths
2. Choose one path
3. Change color of the path
4. Sometimes it will double render the artwork on the stage

CC @cwillisf - @paulkaplan recommended the I cc you on this, as you may have ideas about what is going on with the render.

### LINKED ISSUE #126: Parts of the paint editor are rendered on the Stage
Possibly related to #125, when duplicate costumes are rendered, sometimes parts of the paint editor come with it!

colorpickercraziness

### LINKED ISSUE #132: Revert color fill, too many undo states
I wanted to undo a color fill. However, it took about twenty clicks because each undo would show a color I scrubbed to while picking the color.
from Eric S

### LINKED ISSUE #137: Editing painting causes renderer glitches
As I changed the color of one shape, the whole costume appeared multiple times, in different sizes, on the stage
From Sarah

paulkaplan:
Fixes https://github.com/LLK/scratch-paint/issues/132
Fixes https://github.com/LLK/scratch-paint/issues/125
Fixes https://github.com/LLK/scratch-paint/issues/126
Fixes https://github.com/LLK/scratch-paint/issues/137

This updates the fill and stroke color pickers with logic to only submit the new SVG to the GUI and store an undo state when the popover is closed. This conveniently also fixes the crazy double rendering stuff, and the undo stack aggregation at the same time. 

aggregate-color-updates

The helper `apply*ColorToSelection` was only being used by the fill and stroke color containers, so it is safe to change it to returning whether something has changed instead of calling `onUpdateSvg` directly.

### DIFF
--- a/src/containers/stroke-color-indicator.jsx
+++ b/src/containers/stroke-color-indicator.jsx
@@ -15,9 +15,24 @@ class StrokeColorIndicator extends React.Component {
         bindAll(this, [
             'handleChangeStrokeColor'
         ]);
+
+        // Flag to track whether an svg-update-worthy change has been made
+        this._hasChanged = false;
+    }
+    componentWillReceiveProps (newProps) {
+        const {strokeColorModalVisible, onUpdateSvg} = this.props;
+        if (strokeColorModalVisible && !newProps.strokeColorModalVisible) {
+            // Submit the new SVG, which also stores a single undo/redo action.
+            if (this._hasChanged) onUpdateSvg();
+        } else if (newProps.strokeColorModalVisible && !strokeColorModalVisible) {
+            // If modal is opening, reset the _hasChanged flag to false;
+            this._hasChanged = false;
+        }
     }
     handleChangeStrokeColor (newColor) {
-        applyStrokeColorToSelection(newColor, this.props.onUpdateSvg);
+        // Apply color and update redux, but do not update svg until picker closes.
+        const isDifferent = applyStrokeColorToSelection(newColor);
+        this._hasChanged = this._hasChanged || isDifferent;
         this.props.onChangeStrokeColor(newColor);
     }
     render () {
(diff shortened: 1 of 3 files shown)


towerofnix:
>Fixes
>Fixes
>Fixes
>Fixes

Combo PR! :tada:

src/containers/stroke-color-indicator.jsx
```diff
@@ -15,9 +15,24 @@ class StrokeColorIndicator extends React.Component {
...
+        this._hasChanged = false;
+    }
+    componentWillReceiveProps (newProps) {
+        const {strokeColorModalVisible, onUpdateSvg} = this.props;
+        if (strokeColorModalVisible && !newProps.strokeColorModalVisible) {
+            // Submit the new SVG, which also stores a single undo/redo action.
+            if (this._hasChanged) onUpdateSvg();
+        } else if (newProps.strokeColorModalVisible && !strokeColorModalVisible) {
+            // If modal is opening, reset the _hasChanged flag to false;
+            this._hasChanged = false;
```
fsih: You could put this in the if statement and skip the else if (in both places)

src/helper/style-path.js
```diff
@@ -8,9 +8,9 @@ const MIXED = 'scratch-paint/style-path/mixed';
 /**
  * Called when setting fill color
  * @param {string} colorString New color, css format
- * @param {!function} onUpdateSvg A callback to call when the image visibly changes
+ * @return {boolean} Whether the color application actually changed anything.
```
fsih: Nit: 'Anything' is kind of vague, maybe 'any colors' or 'visibly' (in both places)

fsih (dismissed):
Thanks, that will help a lot!

src/containers/stroke-color-indicator.jsx
paulkaplan: Fixed in 3dab6ef

src/helper/style-path.js
paulkaplan: Fixed in 3dab6ef

fsih (approved):

Issue #167, opened 2017-10-30
Remove Pen tool


carljbowman:
For the next playtest we would like to remove the Pen tool from the Vector Paint Editor to see how that constraint impacts user behavior.

SharkPool-SP:
What the fuck 💀

LTV2008:
??? 🤨🤨🤨

DogeisCut:
society

jwklong:
massive mistake

PR #168, opened 2017-10-30
Remove pen

fsih:
Resolves https://github.com/LLK/scratch-paint/issues/167


TheBrokenRail:
Why not just remove the pen button from ```paint-editor.jsx``` considering this change is just temporary for a play test?

mewtaylor (approved):

Issue #135, opened 2017-10-27
Fill color widget was in an unexpected state (showed multicolor when drawing a circle)


jwzimmer-zz:
Reported by @ericrosenbaum during internal smoke test, 10/27/17 when trying to draw a circle:
> select a group of shapes of different colors, and fill color widget shows "multicolor", then select the circle tool, and fill color widget still shows "multicolor," so I don't know what color it's going to draw it in

PR #169, opened 2017-10-30
Force a color state when switching to shape tools

fsih:
Resolves https://github.com/LLK/scratch-paint/issues/135

If fill and stroke color are both mixed/transparent/absent, set fill to default and stroke to transparent.
If exactly one of fill or stroke color is set, set the other one to transparent.
This way the tool won't draw an invisible state, or be unclear about what will be drawn.

### DIFF
--- a/src/containers/brush-mode.jsx
+++ b/src/containers/brush-mode.jsx
@@ -6,7 +6,7 @@ import Modes from '../modes/modes';
 import Blobbiness from '../helper/blob-tools/blob';
 import {MIXED} from '../helper/style-path';
 
-import {changeFillColor} from '../reducers/fill-color';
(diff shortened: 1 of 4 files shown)


mewtaylor:
@fsih this looks alright to me (though I'm not marked for review so didn't want to approve it without @paulkaplan weighing in).

One thing that I wonder about for the future is if we could/should move some of the common functionality across container components into a base component that can be composed by other components, since they share some similar functionality, but that's certainly not an important discussion at this moment!

paulkaplan (approved):
LG. Agree with matt about extracting this "forcing" logic into a HOC, but it'll take some thinking to figure out the right way to express it.

fsih:
The shape tools definitely share almost all of their component code and should be refactored. The other tools that force state all do it differently though. Maybe it makes sense to leave those in their tools.

PR #170, opened 2017-10-31
Fix GH-131: Add a larger width on small inputs

### LINKED ISSUE #131: Eraser size is cut-off in tool area
from Julia

screen shot 2017-10-27 at 2 19 03 pm

mewtaylor:
Fixes #131 – I ended up deciding to do this for all inputs because the original `3rem` makes it cut off text past 1 char for any input that is only `3rem`, so `3.5rem` seemed like an okay width for this component.

### DIFF
--- a/src/components/forms/input.css
+++ b/src/components/forms/input.css
@@ -41,6 +41,6 @@ See https://github.com/LLK/scratch-paint/issues/13 */
 }
 
 .input-small {
-    width: 3rem; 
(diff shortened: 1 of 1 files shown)


paulkaplan (approved):

fsih:
image
Julia and I still saw this issue at the playtest today (Chrome on Mac and Chrome on Windows 10)

Issue #134, opened 2017-10-27
Add Copy / Paste mode tools


fsih:
I was told you can hold down alt and then move the object, but I would not have known how to do that
from Julia

carljbowman:
![selection tool - 1024x768](https://user-images.githubusercontent.com/3409578/32228576-f976670c-be24-11e7-85ae-b4af1e3605de.png)

[assets.zip](https://github.com/LLK/scratch-paint/files/1430997/assets.zip)

UI States: 
Chatted with @mewtaylor on states
Initial State (Nothing Selected / Nothing Copied)
- Copy: Disabled
- Paste: Disabled

Path Selected / Nothing Copied
- Copy: Enabled
- Paste: Disabled

Path Selected / Path Copied
- Copy: Enabled
- Paste: Enabled

carljbowman:
Adding @mewtaylor to help with layout.

PR #171, opened 2017-10-31
Towards GH-134: Add disabled copy/paste buttons

mewtaylor:
This is for #134 – adds copy/paste buttons to the second row of tools, and keeps them in a disabled state for now, while they’re not usable. It also moves icons into a subfolder, given their proliferation in the paint editor component.

**Without mode tools**
screen shot 2017-10-31 at 1 43 45 pm

**With mode tools**
screen shot 2017-10-31 at 1 44 28 pm

/cc @carljbowman


src/components/paint-editor/paint-editor.jsx
```diff
@@ -205,7 +213,23 @@ class PaintEditorComponent extends React.Component {
                                 <StrokeWidthIndicatorComponent
                                     onUpdateSvg={this.props.onUpdateSvg}
                                 />
-                            </div>
+                            </InputGroup>
+                            <InputGroup className={classNames(styles.modDashedBorder, styles.modLabeledIconHeight)}>
+                                <LabeledIconButton
+                                    disabled
+                                    imgAlt="Copy Icon"
```
fsih: We should define all of these in 'define messages' at some point.

Let's not use the word 'Icon' in the alt text

fsih (changes requested):

src/components/paint-editor/paint-editor.jsx
fsih: I checked it out with a screenreader, and the current state is
- It reads the button and text on the button as separate focus items. Ideally it would be one item
- When it reads the button it seems to randomly pick between saying just "button" and the alt text

Obviously there's a lot to fix for screen readers so I wouldn't spend too much time on it

mewtaylor: Thanks! I just updated with the changes discussed

src/components/paint-editor/paint-editor.jsx
```diff
@@ -50,6 +50,56 @@ const messages = defineMessages({
...
+        id: 'paint.paintEditor.ungroup'
+    },
+    undo: {
+        defaultMessage: 'Undo',
+        description: 'Alt to image for the button to undo an action',
+        id: 'paint.paintEditor.undo'
+    },
+    redo: {
+        defaultMessage: 'Redo',
+        description: 'Alt to image for the button to undo an action',
```
fsih: should be redo

fsih (dismissed):
Looks good besides the one description, thanks!

fsih (approved):

PR #173, opened 2017-11-01
Move copy/paste to mode tools

fsih:
We realized that copy/paste is only designed to work in select mode, so I moved it to mode tools.

I had to duplicate some CSS between paint-editor and mode-tools. Do you have a better idea for how to organize things? Maybe the top buttons should be in their own component outside of paint-editor that shares CSS?


mewtaylor (approved):
@fsih LGTM! In response to your question, I don't know off of the top of my head, but do think there is probably something we can do. Maybe we can take a look after getting stuff together for this week?

PR #174, opened 2017-11-01
Fix GH-122: change size inputs to input with custon onChange

### LINKED ISSUE #122: Change back `BufferedInput` to only update onBlur
The change made to account for the brush sizer and the eraser sizer (so that they would dynamically update the brush on a value change) was really just making them be a ranged input – so we should revert the functionality added to the `BufferedInput` component, and just change those (i.e. outline) to be a regular input.

This affects the the costume name input in an adverse way

mewtaylor:
This fixes #122 by reverting the changes to `bufferedInputHOC`, and adding a custom `onChange` handler to the mode tools for brush and eraser, and making them `Input` components. Same goes for the strok width indicator.

### DIFF
--- a/src/components/forms/live-input-hoc.jsx
+++ b/src/components/forms/live-input-hoc.jsx
@@ -1,0 +1,60 @@
+import bindAll from 'lodash.bindall';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+/**
+ * Higher Order Component to manage inputs that submit on change and <enter>
+ * @param {React.Component} Input text input that consumes onChange, onBlur, onKeyPress
+ * @returns {React.Component} Live input that calls onSubmit on change and <enter>
+ */
+export default function (Input) {
+    class LiveInput extends React.Component {
+        constructor (props) {
+            super(props);
+            bindAll(this, [
+                'handleChange',
+                'handleKeyPress',
+                'handleFlush'
+            ]);
+            this.state = {
+                value: null
+            };
+        }
+        handleKeyPress (e) {
+            if (e.key === 'Enter') {
+                this.handleChange(e);
+                e.target.blur();
+            }
+        }
+        handleFlush () {
+            this.setState({value: null});
+        }
+        handleChange (e) {
+            const isNumeric = typeof this.props.value === 'number';
+            const validatesNumeric = isNumeric ? !isNaN(e.target.value) : true;
+            if (e.target.value !== null && validatesNumeric) {
+                this.props.onSubmit(Number(e.target.value));
+            }
+            this.setState({value: e.target.value});
+        }
+        render () {
+            const liveValue = this.state.value === null ? this.props.value : this.state.value;
+            return (
+                <Input
+                    {...this.props}
+                    value={liveValue}
+                    onBlur={this.handleFlush}
+                    onChange={this.handleChange}
+                    onKeyPress={this.handleKeyPress}
+                />
+            );
+        }
+    }
+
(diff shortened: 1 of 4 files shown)


paulkaplan:
Looks like you can't delete all the text in the box? Once I try to backspace to type a new value, it automatically makes it 0 or 1 (depending on the input).

fsih:
I think the box should let you type whatever you want in it, and while it's invalid the internal state should default to 1 but it shouldn't update the box to say 1, until you unfocus maybe? This way you can make it through the transition state of having an empty field when trying to input a new number

mewtaylor:
@paulkaplan @fsih I just adjusted the PR to have the ability to handle deletion as y'all mentioned through a new HOC. Thoughts?

src/components/forms/live-input-hoc.jsx
```diff
@@ -0,0 +1,60 @@
...
+                this.handleChange(e);
+                e.target.blur();
+            }
+        }
+        handleFlush () {
+            this.setState({value: null});
+        }
+        handleChange (e) {
+            const isNumeric = typeof this.props.value === 'number';
+            const validatesNumeric = isNumeric ? !isNaN(e.target.value) : true;
```
fsih, line 34: I think you need max value checking here. When you type in a huge number and press enter, the value in the field snaps back down to 400 but the image's stroke width doesn't. It looks like for huge values, the canvas just gives up.

src/components/forms/live-input-hoc.jsx
```diff
@@ -0,0 +1,60 @@
...
+            bindAll(this, [
+                'handleChange',
+                'handleKeyPress',
+                'handleFlush'
+            ]);
+            this.state = {
+                value: null
+            };
+        }
+        handleKeyPress (e) {
```
fsih, line 23: Nit: The function names handleChange, handleKeyPress, and handleFlush all kind of sound like the same thing, so it's hard to guess which one will be called by what and what order they'll happen in. Maybe the default wording on an input is just confusing. Do onKeyPress and onChange always both happen? Maybe we can come up with better names?

fsih (changes requested):
Thanks! This HOC looks like it'll be really useful

src/components/forms/live-input-hoc.jsx
mewtaylor: @fsih great point – should we add this for the buffered input too, while I'm at it? /cc @paulkaplan

fsih: It looks like the buffered inputs in the sprite info panel are handling max values OK already?

mewtaylor: @fsih ah ok – odd, since it's the same functionality just in a different method. I'll take a look...

fsih: They're probably handling it outside the component, and it's easier since it doesn't have a "live" factor. But yeah @paulkaplan would know better

mewtaylor:
@fsih I added validation on max input, so that it "snaps" to the max if someone types an amount greater than that – is that the desired functionality?

Also w.r.t. the `handle` naming conventions – I was mostly trying to follow the conventions used already in `bufferedInputHOC`, but could change the names of the methods called if it's confusing for sure!

fsih:
For the naming - that's fine, I think inputs are just confusing.

Sorry for dragging out the review, but I think we should do min validation too. Besides that everything looks good, thanks!

fsih (approved):

PR #180, opened 2017-11-03
Fix null mode issue

fsih:
Guard against right clicks and having no mode in more places. Having no mode breaks the paint editor irrecoverably.


paulkaplan (approved):
LG

PR #185, opened 2017-11-06
Improve cursor preview efficiency

fsih:
Improve cursor preview performance by using a shape, which doesn't have to be recreated after every change in position.


paulkaplan (approved):
LGTM

PR #188, opened 2017-11-06
Widen stroke width indicator

fsih:
Fixes https://github.com/LLK/scratch-paint/issues/131

Happy to hear suggestions on how to improve the placement of this css

4rem comfortably fits the max value of 400
image

### DIFF
--- a/src/components/stroke-width-indicator.css
+++ b/src/components/stroke-width-indicator.css
@@ -1,0 +1,3 @@
+#stroke-width-indicator {
+    width: 4rem;
+}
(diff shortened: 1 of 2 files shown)


mewtaylor:
Ah @fsih sorry I did a PR for this since I thought I was still assigned to it – see #191. I'd recommend that we go with the one that I did to avoid using ids in css like this. How does that sound to you?

fsih:
That sounds great. I don't really know what I'm doing with css

PR #189, opened 2017-11-06
Test adding canvas dep to dev deps

fsih:
Seeing if this fixes Travis


rschamp (changes requested):
If you go this route, you should probably remove the manual installation from the travis install step.

paulkaplan:
Do we want to go this route? It fails on my computer :(

fsih:
Let's not

PR #191, opened 2017-11-07
GH-131 Again: remove `small` from live inputs

### LINKED ISSUE #131: Eraser size is cut-off in tool area

mewtaylor:
Fixes #131 by instead of changing the small width, simply just removing the small tag from things that don’t need it.

### DIFF
--- a/src/components/forms/input.css
+++ b/src/components/forms/input.css
@@ -41,6 +41,6 @@ See https://github.com/LLK/scratch-paint/issues/13 */
 }
 
 .input-small {
-    width: 3.5rem; 
+    width: 3rem; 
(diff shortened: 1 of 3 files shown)


fsih (approved):

fsih:
You mentioned we could use :local to avoid collisions with GUI. That sounds like a good idea. Could we do that instead, or were there other issues with it?

mewtaylor:
@fsih we could, for sure – however when I was looking, it just seemed like it was unnecessary to introduce that when we could just remove the `small` tag

Issue #41, opened 2017-09-26
Paint editor minimum width UI


fsih:
When the screen is below a certain width, switch to the more compact version of the UI

mewtaylor:
fixed in #194

PR #194, opened 2017-11-07
Implement GH-41: Add minimum width gui

mewtaylor:
This implements #41. It adds a new component, `Dropdown`, which is a formatted version of the `Popover` element.

I also moved the `modes.js` file to a `lib` folder, and added `layout-constants` from gui to the lib folder as well. This move from `modes/modes.js` to `lib/modes.js` can be reverted if needed.

@carljbowman and I discussed using the `ContextMenu` for this, however that library is pre-built to pop up wherever there is a click, which is not the desired behavior for this feature.

We could feasibly move the styling for `.mod-menu-item` into its own component on dropdown if we want, but I wasn't sure if that was a good idea or not.


package.json
```diff
@@ -67,6 +67,7 @@
     "react-intl-redux": "0.6.0",
     "react-popover": "0.5.4",
     "react-redux": "5.0.5",
+    "react-responsive": "^3.0.0",
```
rschamp: We should pin this since it's actively used by the application

src/lib/layout-constants.js
```diff
@@ -0,0 +1,3 @@
+export default {
+    fullSizeMinWidth: 1096
```
fsih: This one says 1096 and the other one says 1095. Is that right?

src/components/paint-editor/paint-editor.jsx
```diff
@@ -342,9 +402,7 @@ PaintEditorComponent.propTypes = {
     canUndo: PropTypes.func.isRequired,
     intl: intlShape,
     name: PropTypes.string,
-    onCopyToClipboard: PropTypes.func.isRequired,
     onGroup: PropTypes.func.isRequired,
-    onPasteFromClipboard: PropTypes.func.isRequired,
```
fsih: I think you want to add these back too

src/components/paint-editor/paint-editor.jsx
```diff
@@ -248,7 +308,7 @@ class PaintEditorComponent extends React.Component {
                             </InputGroup>
                             <InputGroup className={styles.modModeTools}>
                                 <ModeToolsComponent
-                                    onCopyToClipboard={this.props.onCopyToClipboard}
+                                    onCopyToClipboard={this.props.onCopyToClipboard}
```
fsih: Nit: there are some extra spaces here

src/lib/layout-constants.js
fsih: Also noticed that things stick out a bit before min width UI kicks in, so we might want to do it sooner?
image

fsih (changes requested):

fsih:
If the 1096/1095 difference is intentional, then this PR looks good to me

fsih (dismissed):

paulkaplan:
@fsih I think this requires your re-review.

fsih (approved):
Sorry! Thanks for the reminder ping

Issue #140, opened 2017-10-27
Make lines the right thickness


fsih:
When you're zoomed in, line thickness is not accurately represented. Line thickness doesn't change with zoom.

jwzimmer-zz:
Closing https://github.com/LLK/scratch-paint/issues/139 as a dupe of this issue:

I didn't change the line width when I was working on this sprite, but you can see how the apparent line width changed over time as I zoomed in & out repeatedly:
<img width="464" alt="screen shot 2017-10-27 at 3 01 45 pm" src="https://user-images.githubusercontent.com/1320456/32122597-e4e40c0c-bb2e-11e7-8e03-bd2f6574b25e.png">
<img width="476" alt="screen shot 2017-10-27 at 3 52 30 pm" src="https://user-images.githubusercontent.com/1320456/32122595-e4cb63b4-bb2e-11e7-863a-c055a481c58d.png">
<img width="432" alt="screen shot 2017-10-27 at 3 52 15 pm" src="https://user-images.githubusercontent.com/1320456/32122596-e4d61322-bb2e-11e7-8ae8-b6e1e6f9d9be.png">

paulkaplan:
I can't seem to get a consistent repro of this, but I know it happened to me at the playtest. Can you lay it out step-by-step?

jwzimmer-zz:
@paulkaplan, I was trying to repro this, and it's not clear to me what the steps are yet (sorry) but I did see it again. I wonder if it has to do with that I'm both setting the zoom for Chrome & for the editor (I have to zoom out in Chrome in order to see the zoom buttons in the editor)?
<img width="1196" alt="screen shot 2017-10-30 at 1 11 16 pm" src="https://user-images.githubusercontent.com/1320456/32185254-069201c2-bd75-11e7-86d1-c84533523718.png">

fsih:
I figured out the repro

1. Select and scale anything
2. That thing now always has the same stroke widths when zoomed in

PR #195, opened 2017-11-07
Clean up the scale tool

fsih:
- Scale tool was setting `strokeScaling` of whatever it touched to false. This was causing https://github.com/LLK/scratch-paint/issues/140.
- Once strokeScaling is added back in, it is apparent that setting `applyMatrix = false` had unwanted side effects. Strokes would appear to scale and then snap back to expected look on mouse up. This also caused an issue with how shapes with gradient fills were previewed (the gradient would not move with the shape). To fix this I added applyMatrix back in and changed how we apply scale transforms accordingly (keeping track of the current amount of transform so that we can apply the delta instead of the entire transform each time)
- Scale tool doesn't need access to bounding box tool internals because the bounding box is removed when the scale tool is activated (see comment "`// while transforming, don't show bounds`" in bounding box tool). Thus a lot of the code was unneeded because it did nothing
- `removeBoundsPath` was removing all helper items, not just bounds. The other helper items included the group that the scale tool makes to show a preview of the change (so the object we're scaling would disappear between mouse down and up). Fixed this to only remove bounds.
- Scale tool was recreating the group that was being scaled in every call to `onMouseDrag` (probably because it didn't work in `onMouseDown` because of the above bug.) Moved this to create it once in `onMouseDown`.

### DIFF
--- a/src/helper/guides.js
+++ b/src/helper/guides.js
@@ -84,8 +84,8 @@ const _removePaperItemsByTags = function (tags) {
     }
 };
 
-const removeHelperItems = function () {
-    _removePaperItemsByDataTags(['isHelperItem']);
+const removeBoundsPath = function () {
+    _removePaperItemsByDataTags(['isSelectionBound', 'isRotHandle', 'isScaleHandle']);
 };
 
 const removeAllGuides = function () {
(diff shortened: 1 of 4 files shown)


mewtaylor (approved):

fsih:
Not sure how you found this PR but thanks Matt!

Issue #193, opened 2017-11-07
`Object.values` is not widely supported. 


paulkaplan:
We should use something other than `Object.values`, as it isn't supported within 2 versions of safari.  We could also use a polyfill. Cause of https://github.com/LLK/scratch-gui/issues/863

PR #196, opened 2017-11-08
Do not use Object.values

paulkaplan:
This is almost silly because the "keymirror" used to create the `Modes` object
actually ensures that the keys and the values of the object are the
same ([source](https://github.com/STRML/keyMirror/blob/master/index.js#L44-L48)). So replace `Object.values` with `Object.keys` which has complete
browser support.

Fixes https://github.com/LLK/scratch-gui/issues/863
Fixes https://github.com/LLK/scratch-paint/issues/193


fsih (approved):

fsih:
Yes very in go

fsih:
Yes very in go

fsih:
Yes very in go

Issue #207, opened 2017-11-30
Import SVG should be wrapped in a try/catch (for watermelons)


paulkaplan:
Some SVG's will invariably break the paint editor (for example, the watermelon sprite right now), so the svg loading process should probably be defensive and be wrapped in a try/catch to make sure it doesn't blow up the gui. 

Repro:
- import watermelon sprite and open the costume editor.

PR #208, opened 2017-11-30
Handle null items

fsih:
Fixes https://github.com/LLK/scratch-paint/issues/207

The watermelon doesn't get imported, it shows a blank editor. But it doesn't destroy everything now.


paulkaplan (approved):

paulkaplan:
Cool. We should file another issue for actually fixing those sprites / figuring out why they do not import before merging this, so that it is tracked.

fsih:
Filed https://github.com/LLK/scratch-paint/issues/209

Issue #147, opened 2017-10-27
Add Eyedropper to the Color Picker


jwzimmer-zz:
@fsih reported:
> Need to be able to select same color

She was drawing a purple alpaca and having trouble finding the same color of purple multiple times.

<img width="400" alt="screen shot 2017-10-27 at 15 39 34" src="https://user-images.githubusercontent.com/1320456/32123335-c53f0660-bb31-11e7-80ca-d0944bbd6f87.png">

carljbowman:
Sketch:
![selection tool - transparent - 1024x768](https://user-images.githubusercontent.com/3409578/32243700-7e042550-be4c-11e7-857d-ae87721c95ed.png)

Assets:
[eye-dropper-icon.zip](https://github.com/LLK/scratch-paint/files/1431959/eye-dropper-icon.zip)

Question:
@fsih @paulkaplan @rschamp - Depending on how you all are thinking about the eye dropper from a technical perspective, we could do a couple things to indicate the eyedropper is selected and ready to use: 
1. Change the cursor (when over the canvas) to the "loop" element that @paulkaplan built for picking color from the stage (triggered on the Touching [ Color ] block).
2. Change the cursor (when over the canvas) to an eye-dropper icon.
3. Open to other suggestions...

Let me know if you all have thoughts.

carljbowman:
Assigning over to @mewtaylor 

Sketches based on our discussion with @paulkaplan @fsih yesterday: https://invis.io/HDDDQ2P7W

mewtaylor:
Implemented in #212

PR #212, opened 2017-12-08
Implement eye dropper for paint editor

mewtaylor:
This implements #147.

**Expected Behavior:**
* eye dropper should hide when the mouse hovers off of the canvas
* eye dropper should deactivate when clicking outside the canvas
* items selected before activating the eye dropper should stay selected during eye dropping
* choosing the background does not choose transparent, it chooses white or gray

This also splits the `ColorPicker` into container/presentation.

@fsih @paulkaplan I'm down to review this together next week if we want, in person!


paulkaplan:
@mewtaylor there are a few lint failures here

paulkaplan:
Also, I'm getting an error when trying to load this in the GUI. Are you seeing the same thing?

```
Uncaught Error: Element ref was specified as a string (wrappedInstance) but no owner was set. You may have multiple copies of React loaded.
```

src/components/box/box.css
```diff
@@ -0,0 +1,2 @@
+.box {
```
fsih: What does this do?

src/components/loupe/loupe.jsx
```diff
@@ -0,0 +1,108 @@
...
+import React from 'react';
+import bindAll from 'lodash.bindall';
+
+import Box from '../box/box.jsx';
+
+import {LOUPE_RADIUS, CANVAS_SCALE} from '../../helper/eye-dropper';
+
+import styles from './loupe.css';
+
+const zoomScale = 3;
```
fsih: Nit: capital constants

src/containers/paint-editor.jsx
```diff
@@ -38,14 +40,37 @@ class PaintEditor extends React.Component {
...
+            'stopEyeDroppingLoop'
         ]);
+        this.state = {
+            colorInfo: null
+        };
     }
     componentDidMount () {
         document.addEventListener('keydown', this.props.onKeyPress);
     }
+    shouldComponentUpdate (nextProps, nextState) {
```
fsih: Did you need to add this to optimize for something in particular?

src/containers/paint-editor.jsx
```diff
@@ -145,12 +170,58 @@ class PaintEditor extends React.Component {
...
+            if (!this.eyeDropper.hideLoupe) {
+                // If not hide loupe, that means the click is inside the canvas,
+                // so apply the new color
+                callback(colorString);
+            }
+            this.setState({colorInfo: null});
+        }
+    }
+    startEyeDroppingLoop () {
+        const canvas = this.paintEditor.getWrappedInstance().canvas;
```
fsih: Instead of doing this, can you get the canvas directly by passing componentRef as a prop and having paint editor set the canvas? getWrappedInstance seems a bit hacky.

src/containers/paint-editor.jsx
```diff
@@ -145,12 +170,58 @@ class PaintEditor extends React.Component {
...
+            this.stopEyeDroppingLoop();
+            if (!this.eyeDropper.hideLoupe) {
+                // If not hide loupe, that means the click is inside the canvas,
+                // so apply the new color
+                callback(colorString);
+            }
+            this.setState({colorInfo: null});
+        }
+    }
+    startEyeDroppingLoop () {
```
fsih, line 187: paint-editor.jsx is getting kind of huge. I wonder if there's a good way to split code like this into other components or helper files before it gets unmanageable.

src/helper/eye-dropper.js
```diff
@@ -0,0 +1,71 @@
+import paper from '@scratch/paper';
+
+const PAPER_WIDTH = 864;
+const PAPER_HEIGHT = 648;
```
fsih: Where do these numbers come from?

src/helper/eye-dropper.js
```diff
@@ -0,0 +1,71 @@
+import paper from '@scratch/paper';
```
fsih: This file might make sense in helper/tools

src/reducers/eye-dropper.js
```diff
@@ -0,0 +1,55 @@
+const ACTIVATE_EYE_DROPPER = 'scratch-paint/eye-dropper/ACTIVATE_COLOR_PICKER';
+const DEACTIVATE_EYE_DROPPER = 'scratch-paint/eye-dropper/DEACTIVATE_COLOR_PICKER';
+
+const initialState = {
```
fsih, line 4: Maybe this state should be part of the mode state?

src/helper/eye-dropper.js
```diff
@@ -0,0 +1,71 @@
...
+const CANVAS_SCALE = 1.8;
+
+class EyeDropperTool extends paper.Tool {
+    constructor (canvas) {
+        super();
+
+        this.onMouseDown = this.handleMouseDown;
+        this.onMouseMove = this.handleMouseMove;
+
+        this.active = false;
```
fsih: Looks like this is unused?

src/helper/eye-dropper.js
```diff
@@ -0,0 +1,71 @@
...
+        this.canvas = canvas;
+        this.colorInfo = null;
```
fsih: This too

src/helper/eye-dropper.js
```diff
@@ -0,0 +1,71 @@
...
+    handleMouseDown () {
+        if (!this.hideLoupe) {
+            const colorInfo = this.getColorInfo(this.pickX, this.pickY, this.hideLoupe);
+            const r = colorInfo.color[0];
+            const g = colorInfo.color[1];
+            const b = colorInfo.color[2];
+
+            const componentToString = c => {
+                const hex = c.toString(16);
+                return hex.length === 1 ? `0${hex}` : hex;
```
fsih: Can you add a comment to explain this bit?

fsih (changes requested):
Nice!

src/components/box/box.css
mewtaylor: it was a copy directly from gui. I can remove it – I thought it might be good to mirror gui more directly, but that seems unnecessary with this

src/containers/paint-editor.jsx
mewtaylor: for optimizing the `setInterval` method to not re-render every time the color info is changed

src/containers/paint-editor.jsx
mewtaylor: I think I'm just going to use `document.getElement` rather than either, due to another bug I'm running into (a bug with refs that @paulkaplan mentions above)

src/helper/eye-dropper.js
mewtaylor: I had to look at the rendered canvas to see them – I can't find a reference anywhere in our code, but this appears to be something that paper is doing, and this is necessary to scale the loupe to make sure it's looking at the right part of the canvas when extracting a color.

src/containers/paint-editor.jsx
mewtaylor: I agree that it's getting huge, and that it'd be good to split things up. The reason I had this be in here rather than elsewhere is mostly to handle the context of the `Loupe` component. This component needs to exist at the context of the paint editor – rather than at the context of the canvas, for example – because we defined the behavior of it as hiding/going away on hover/click outside of the canvas itself. 

Given this issue, it seemed necessary to me for the component to exist at the paint editor level rather than in some separate component. It probably can be moved out, though, if that's a dealbreaker.

If we're looking to make the paint editor component smaller, though, I think we could get some easier refactors out of breaking out `Edit Tools` and `Mode Tools` into their own components.

src/reducers/eye-dropper.js
mewtaylor: I'd leave that up to you and @paulkaplan if you think it makes sense – moving this to `reducers/modes.js` would mean adding additional functionality to the mode state, since right now, across all modes, it simply changed the name of the mode. Including more stuff in there would be new, and seem to only apply to the eye dropper. If that's cool with you two, then I can do that for sure

mewtaylor:
@paulkaplan should work now, thanks for seeing that – it was an issue with `react-intl` and its `withRef` argument in a library component somehow. And @fsih I made some of your requested changes and left comments for the others, let me know if you'd like to see any other changes based on what's here!

src/components/box/box.css
fsih: Let's take it out if it's not doing anything

src/helper/eye-dropper.js
fsih: image

The coordinates are really off for me. We'll probably have to figure out where these numbers are coming from

src/reducers/eye-dropper.js
fsih: There's 2 places to look for whether the eyedropper is active now, this.active and mode = eyedropper, and I'm worried that they could get out of sync if we forget about one or the other. Alternatively, what if we don't change the mode when switching to the eyedropper? Are there issues with another tool mode being active while the eyedropper is running?

src/containers/paint-editor.jsx
fsih: I think it's discouraged to use document.getElement in react? Paul would know better. Maybe we can figure out the ref bug together without using document.getElement.

src/containers/paint-editor.jsx
fsih: That makes sense. Let's file a bug to break those out when we have time.

src/containers/paint-editor.jsx
fsih: It's checking this.state.colorInfo !== nextState.colorInfo so wouldn't it still rerender?

src/reducers/eye-dropper.js
mewtaylor: There are issues with another tool mode being active while the eye dropper is running – this was causing issues with mouse events, which was why we moved this to being a tool, rather than having its own mouse events. 

I'm fine with moving the data that's needed for the eye dropper into the mode itself as a resolution to the issue of it being in two places, but for a reason I can't seem to pin down, just activating the eye dropper as a tool without deactivating another first disables that tool when the eye dropper is then deactivated after it finishes.

Alternatively – I could move the "isActive" logic to just use the `mode`, but keep the eye dropper info logic in the separate reducer. That way, the activation logic is confined to one place, but the data is not cluttering up the `modes` reducer. Thoughts @fsih ?

src/containers/paint-editor.jsx
mewtaylor: This handles the case of a) hovering outside of the confines of the canvas, or b) if the loupe is paused in a particular place for a while. If we're not too concerned with performance in this case, then I can take it out though

src/containers/paint-editor.jsx
mewtaylor: @fsih I changed it in a different way that prevents using `getElement` or `getWrappedInstance` – what do you think of the way it is now, with the canvas being a property rather than a state variable?

src/containers/paint-editor.jsx
```diff
@@ -200,10 +278,14 @@ PaintEditor.propTypes = {
...
 
 const mapStateToProps = state => ({
-    selectedItems: state.scratchPaint.selectedItems,
-    undoState: state.scratchPaint.undo,
+    changeColorToEyeDropper: state.scratchPaint.color.eyeDropper.callback,
     clipboardItems: state.scratchPaint.clipboard.items,
-    pasteOffset: state.scratchPaint.clipboard.pasteOffset
+    isEyeDropping: state.scratchPaint.color.eyeDropper.active,
+    pasteOffset: state.scratchPaint.clipboard.pasteOffset,
+    previousItems: state.scratchPaint.color.eyeDropper.previousItems,
```
fsih: Is this left over?

src/reducers/eye-dropper.js
fsih: Re: the 2nd paragraph, after you deactivated the eye dropper, it might be that you have to call tool.activate() on the old tool.

(But don't worry about it since I think the other way we were discussing is probably better anyway)

src/containers/paint-editor.jsx
fsih: Can we check the colorInfo equality before we call setState instead of skipping the rendering in shouldComponentUpdate?

src/helper/tools/eye-dropper.js
```diff
@@ -0,0 +1,68 @@
+import paper from '@scratch/paper';
+
+const LOUPE_RADIUS = 20;
```
fsih, line 3: Nice! This looks waaay better

fsih:
I found a couple of bugs.

1. When you have something selected and you pick a color, an undo state doesn't get set/ the stage doesn't get updated. The color picker currently doesn't save undo state every time the color changes, but it should save it when it gets closed. So we should make sure FillColorIndicator's handleChangeFillColor gets called (to set the _hasChanged flag) and then componentWillReceiveProps gets called when it's closed (so that it calls onUpdateSvg)

2. When you zoom in, the loupe movement gets off:
mouse offset
(the zoom ratio is in paper.view.zoom)

mewtaylor:
@fsih hmm – `handleChangeFillColor` is what's called, as it's what is passed in as the `callback` value in redux, so `_hasChanged` should be getting changed... I'll look to see if there's a bug with that. And thanks for pointing out the zoom thing!

src/containers/color-picker.jsx
```diff
@@ -35,25 +35,37 @@ class ColorPicker extends React.Component {
...
+        const hsv = this.getHsv(props.color);
         this.state = {
             hue: hsv[0],
             saturation: hsv[1],
             brightness: hsv[2]
         };
     }
-    componentWillReceiveProps () {
-        // Just a reminder, new props do not update the hsv state
+    componentWillReceiveProps (newProps) {
```
fsih: You should probably get @paulkaplan to check this color picker stuff

fsih (dismissed):
Awesome! LGTM

paulkaplan:
@mewtaylor looks great! A couple minor things.
1. With the stage picker, the cursor is hidden and the central square is filled in with the picked color. I think we should keep this consistent with that. If you want I can contribute this change directly to your branch, there is a minor complication because you can't use `cursor: none` on something with `pointer-events: none`, so it isn't as simple as a single line of CSS. It'll require adding a class to the canvas container (since that can receive pointer events).
2. The eyedropper icon is... extremely tiny. We should probably bump the size of that and the swatch to at least a touchable 40x40. That was my bad originally for making that transparent swatch so tiny. /cc @carljbowman 
3. I get a console error when the eyedropper starts up about `Warning: `NaN` is an invalid value for the `top` css style property.` That may or may not be related to...
4. The initial loupe state seems strange, when you click it an empty loupe is shown at the top left...
initial-loupe-position

src/components/loupe/loupe.jsx
```diff
@@ -0,0 +1,110 @@
...
+                width={LOUPE_RADIUS * 2}
+            />
+        );
+    }
+}
+
+LoupeComponent.propTypes = {
+    colorInfo: PropTypes.shape({
+        color: PropTypes.shape({
+            r: PropTypes.number,
```
paulkaplan: Looks like there is a disagreement about the color format. In the proptypes and line 55, it assumes `{r: r, g: g, b: b, a: a}`, but in line 60 here and in the eye-dropper-tool `getColorInfo` it assumes `[r, g, b, a]`. I don't have a preference either way, but since this component should ideally match the GUI version, I think we should use the object style.

paulkaplan (changes requested):

paulkaplan (approved):

paulkaplan:
Looks good to me! I did some testing on this integrated with the GUI, but let's make sure to emphasize this in our next internal test.

Issue #209, opened 2017-12-01
Fix the watermelon sprite (and check for others)


fsih:
The watermelon sprite isn't importing.

paulkaplan:
It seems to be that paperjs cannot import SVGs that use the `<svg:` namespace. For example, the watermelon costume for some reason namespaces all of the elements with `svg:`. it seems that inkscape did this. You can also see it in projects like https://llk.github.io/scratch-gui/develop/#177260519 where many costumes do not import.

Seems like a paperjs bug, because the svg does declare an `svg` namespace correctly (xmlns:svg="http://www.w3.org/2000/svg"), and chrome/preview both support it. But as a simple workaround, a string replacement before importing fixes this (`svg.split('svg:').join('')`) does fix it.

paulkaplan:
Not sure if you get notifications if I don't tag you, so for completeness, @fsih ^ :)

paulkaplan:
Also, wrote a script to check all the other costumes in the library. Only watermelon a, b and c suffer from that svg: namespace pathology.

PR #213, opened 2017-12-11
Remove "svg:" namespace from svg before importing.

paulkaplan:
The namespace on tags causes paperjs to fail to import.

Fixes https://github.com/LLK/scratch-paint/issues/209

### DIFF
--- a/src/containers/paper-canvas.jsx
+++ b/src/containers/paper-canvas.jsx
@@ -81,6 +81,19 @@ class PaperCanvas extends React.Component {
     }
     importSvg (svg, rotationCenterX, rotationCenterY) {
         const paperCanvas = this;
+        // Pre-process SVG to prevent parsing errors (discussion from #213)
+        // 1. Remove newlines and tab characters, chrome will not load urls with them.
+        //      https://www.chromestatus.com/feature/5735596811091968
+        svg = svg.split(/[\n|\r|\t]/).join('');
+        // 2. Remove svg: namespace on elements.
+        svg = svg.split(/<\s*svg:/).join('<');
+        svg = svg.split(/<\/\s*svg:/).join('</');
+        // 3. Add root svg namespace if it does not exist.
+        const svgAttrs = svg.match(/<svg [^>]*>/);
+        if (svgAttrs && svgAttrs[0].indexOf('xmlns=') === -1) {
+            svg = svg.replace(
+                '<svg ', '<svg xmlns="http://www.w3.org/2000/svg" ');
+        }
         paper.project.importSVG(svg, {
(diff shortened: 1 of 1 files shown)


fsih (commented):
Should we replace it with a different namespace, in case it was namespaced to avoid conflicts? Do all namespaces break paper?

paulkaplan:
I'm not sure, this really is a paperjs bug, but I thought this would be an ok way to fix it for now.

fsih:
We'll probably be making do with just this fix for a long time, unless we want to put high priority on fixing the paper.js library bug, which has pretty complicated import code. Inkscape probably added namespacing because something they use collides with something in svg. So I feel like we should replace 'svg:' with something, or something may break in a hard-to-find way when someone imports an SVG made in inkscape in the future.

paulkaplan:
@fsih here is the reason that svg namespace (and in fact, all namespaces) will be a problem with paper. After parsing the svg, it "moves" the nodes into a new svg element without transferring any of the namespaces (see [here](https://github.com/paperjs/paper.js/blob/develop/src/svg/SvgImport.js#L584-L589) and [here](https://github.com/paperjs/paper.js/blob/6e7fe67a8ddc93d41b620ef8f02a503d2920769b/src/svg/SvgElement.js)). It does use the svg namespace as the root namespace though, so it should be safe to replace all the `svg:` with nothing. We couldn't replace them with a different namespace for the same reason (we do not have control over what namespaces are loaded). 

I modified the processing to force add a default namespace for svg. This just makes sure that the dom parsing succeeds if the svg had been using `svg:` namespacing without a default namespace (strangely the watermelon actually declared both a default namespace and the svg: namespace).

fsih (dismissed):
Thanks for explaining XML namespace stuff to me!

src/containers/paper-canvas.jsx
```diff
@@ -81,7 +81,14 @@ class PaperCanvas extends React.Component {
     }
     importSvg (svg, rotationCenterX, rotationCenterY) {
         const paperCanvas = this;
-        paper.project.importSVG(svg, {
+        let processedSvg = svg.split('svg:').join('');
+        if (processedSvg.indexOf('xmlns=') === -1) {
```
towerofnix: Is there any chance this could match, say, an SVG with a text element that contains "xmlns="?

Scratch 2.0 paint editor including "xmlns=" text

paulkaplan: Haha good catch. I rewrote the process to only check within the svg tags attributes. @fsih can you re-review.

fsih (approved):

Issue #211, opened 2017-12-07
Increase size of bounding box handles


paulkaplan:
Feedback from playtest. @carljbowman mentioned there is a spec for this in invision.

carljbowman:
Here is an initial sketch. We are balancing making these larger so they are easier to click and not having too big of UI that obscures the artwork. In the sketch below the handles are 8px by 8px, with a 2px drop-shadow. 

![selection tool - 1024x768](https://user-images.githubusercontent.com/3409578/33733735-a965ae66-db58-11e7-9eee-bd94875e9143.png)

Obviously we will need to come up with an alternative solution on touch devices.

paulkaplan:
@fsih is this a straightforward fix? I'd love to have this in before our next workshop if possible.

fsih:
Probably. When is the next workshop?
Those drop-shadows look like translucent circles? What color/opacity are they?

carljbowman:
@fsih - The drop shadows in the sketch are the motion color at 25% opacity. We can dial that in if its too much or too little.

motion primary HEX: #4C97FF (has not opacity)
hsla(215, 100%, 65%, 0.25)

PR #214, opened 2017-12-12
Selection box handles

fsih:
Redesign the selection box in accordance with https://github.com/LLK/scratch-paint/issues/211#issuecomment-350755283

handles

### DIFF
--- a/src/helper/hover.js
+++ b/src/helper/hover.js
@@ -11,6 +11,7 @@ import {isGroupChild} from './group';
  * @return {paper.Item} the hovered item or null if there is none
  */
 const getHoveredItem = function (event, hitOptions, subselect) {
+    // @todo make hit test only hit painting layer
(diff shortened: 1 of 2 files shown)


src/helper/hover.js
```diff
@@ -11,6 +11,7 @@ import {isGroupChild} from './group';
  * @return {paper.Item} the hovered item or null if there is none
  */
 const getHoveredItem = function (event, hitOptions, subselect) {
+    // @todo make hit test only hit painting layer
```
paulkaplan, line 14: ?

paulkaplan (approved):
Changes look good, curious about the todo comment though.

src/helper/hover.js
fsih: I added this todo because we have to set data.noHover and data.noSelect all over the place when it should just be the default for everything that's not on the painting layer. It would make the code cleaner.

PR #221, opened 2017-12-14
Fill tool

fsih:
fillingfills

Filling fills! No filling strokes yet.


paulkaplan:
Testing this manually I'm seeing some strange behavior with hover. Watch how sometimes when i leave the canvas to the left and then come back (or just move in and out of the rect quickly), it doesn't show the rectangle as being hovered. But when I go and hover other cat parts and come back, it seems to hover/fill just fine. 

hover-behaviour

paulkaplan:
Woops, looks like that was because of the transparent hit change to paper, I wasn't updated on my dependencies locally. Nevermind that.

src/helper/math.js
```diff
@@ -27,9 +27,79 @@ const snapDeltaToAngle = function (delta, snapAngle) {
...
+            return null;
+        }
+        temp = temp.parent;
+    }
+    return depth;
+};
+
+const sortItemsByZIndex = function (a, b) {
+    if (a === null || b === null) {
+        // Incomparable
```
paulkaplan: _/warning: irrelevant language rabbit hole follows_

I stumbled on "incomparable" in this context, generally it means "something that is so good that it is beyond comparison". However, dictionaries don't (yet) recognize "uncomparable" as a word, but there isn't really a good word that unambiguously means "cannot be compared" (although when speaking I think the emphasis changes depending on the meaning). I guess "incommensurable" means that, but it is a mouthful and has more of a philosophical bent than is necessary here, and it is easily mistaken for "incommensurate", which is completely different.

rschamp: "Not comparable" can also work although it's two words, I counted.

paulkaplan: would you say "not COMPrable" or "not comPAIRable"?

I feel like saying "inCOMPrable" means beyond compare, and "incomPAIRable" means it doesn't makes sense/is impossible to compare.

fsih: image

fsih:
@paulkaplan I took out grouping of filled things, so fill-tool should be good to review now.

paulkaplan (dismissed):
So this is pretty complex, but I did my best to trace all the deep property accessing to make sure it was all kosher and it seems to be ok. I also tried pretty hard to break it, and wasn't able to. LGTM!

src/helper/math.js
paulkaplan: But that example sentence is super ambiguous! What does it even mean? Is censorship worse now, or less bad?!

fsih:
I broke out the long boolean you mentioned

Issue #218, opened 2017-12-13
Selection bounding box disappears while resizing / rotating


carljbowman:
**Currently on Develop**
Browser: Chrome 63.0.3239.84
OS: Mac OS X 10.11.6
![bounding box](https://user-images.githubusercontent.com/3409578/33964696-4d481654-e027-11e7-9db5-f16ae1b25868.gif)

1. Select an object
2. Scale or rotate the object
3. Bounding box UI disappears

**Expected Behavior**
The bounding box should remain visible and resize with the artwork.

PR #222, opened 2017-12-15
Show bounding box when moving and scaling in the select tool

fsih:
bounds

Fixes https://github.com/LLK/scratch-paint/issues/218

### DIFF
--- a/src/helper/guides.js
+++ b/src/helper/guides.js
@@ -88,6 +88,10 @@ const removeBoundsPath = function () {
     _removePaperItemsByDataTags(['isSelectionBound', 'isRotHandle', 'isScaleHandle']);
 };
 
+const removeBoundsHandles = function () {
+    _removePaperItemsByDataTags(['isRotHandle', 'isScaleHandle']);
+};
+
 const removeAllGuides = function () {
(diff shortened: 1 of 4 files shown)


paulkaplan:
@fsih would it be possible to update the positions of the bounds handles instead of removing them?

fsih:
@paulkaplan I thought we agreed to just put in the box for now?

We actually used to show the handles when scaling, but it was really messy. We had to pass all of that UI from the bounding box tool to the scale and move tools for them to control, and I feel like the bounding box tool alone should be in charge of what the bounding box looks like ideally. Coming up with a good solution for this would take some time, and this bug is technically not in the Dec 20 iteration, so I wanted to put that off for now.
https://github.com/LLK/scratch-paint/commit/fd9a4af83f859565ed4c1e37993592c4f9bf2bb8#diff-e753d3454a0c89ede7c07efb1cc2b3b7L106

paulkaplan:
@fsih ok i must have misunderstood. I thought we were going to move the handles and the box for everything except the rotation tool. LGTM

paulkaplan (approved):
LGTM

PR #229, opened 2017-12-19
Add missing code for updating stroke color from eyedropper

paulkaplan:
@mewtaylor looks like this code was missing from that eyedropper fix you made before the PR was merged. I just copied it over from the fill color container. 

Here is what it looks like on develop (eyedropper doesn't update the stage)
not-working-stroke-eyedropper

Here is what it does after the change (updates the stage) 
working-stroke-eyedropper


mewtaylor (approved):
Ah right – forgot to apply the same thing in fill and stroke. Thanks @paulkaplan :)

Issue #227, opened 2017-12-18
Issue filling dog


fsih:
On the https://llk.github.io/scratch-gui/paint-fill-test-branch/
When you try to fill the patch on the back of the dog, it will only fill the outline:
![image](https://user-images.githubusercontent.com/2855464/34124270-3df2dda6-e400-11e7-915c-068e82e2c460.png)

rschamp:
I also saw this issue with the body of Dinosaur1-c:

![image](https://user-images.githubusercontent.com/1512021/34124888-8e3e23a4-e402-11e7-82c1-bdc1a8919ef8.png)

fsih:
The issue is that the compound paths aren't clockwise.

PR #232, opened 2017-12-19
Convert SVGs to clockwise when importing

fsih:
Fixes https://github.com/LLK/scratch-paint/issues/227

When importing SVGs, convert them to clockwise. That is for a compound path, the largest path item will have positive area, and holes will have negative area.


paulkaplan (approved):
Wow I wish more things in life worked this way. 

```
thesis.written = true;

mortgage.paidOff = true;
```

Issue #233, opened 2017-12-19
If you scale anything, the bounding box appears on the stage


fsih:
The bounding box should not appear on the stage

PR #237, opened 2017-12-19
Remove the bounds path when you finish scaling

fsih:
Fixes https://github.com/LLK/scratch-paint/issues/233

(Adding 'active' is a safety check, since I was able to get onMouseDown to happen multiple times before onMouseUp by using right click and left click together)


paulkaplan (approved):
LG

PR #238, opened 2017-12-20
Don't change opposite handle length when changing handle length

fsih:
Before, you would get stuck in situations like this, where it's very hard to manipulate a curve with uneven handles:
handle
After:
handle2

/cc @carljbowman


paulkaplan (approved):
LG

PR #242, opened 2017-12-21
Toggle curved/pointed points in reshape tool

fsih:
Fixes https://github.com/LLK/scratch-paint/issues/22

curved
curvycat

Note that this makes the selected-items reducer more lax in how it checks for changes. That's because when toggling pointed/curved, it doesn't actually change the selected items, only the handles in the selected items, and to check equality to that depth I would have to do a lot of duplicating objects.

### DIFF
--- a/src/containers/mode-tools.jsx
+++ b/src/containers/mode-tools.jsx
@@ -13,21 +13,115 @@ class ModeTools extends React.Component {
     constructor (props) {
         super(props);
         bindAll(this, [
-            'hasSelectedPoints',
+            '_getSelectedUncurvedPoints',
+            '_getSelectedUnpointedPoints',
+            'hasSelectedUncurvedPoints',
+            'hasSelectedUnpointedPoints',
             'handleCopyToClipboard',
-            'handlePasteFromClipboard'
+            'handleCurvePoints',
+            'handlePasteFromClipboard',
+            'handlePointPoints'
         ]);
     }
-    hasSelectedPoints () {
+    _getSelectedUncurvedPoints () {
+        const items = [];
         const selectedItems = getSelectedLeafItems();
         for (const item of selectedItems) {
+            if (!item.segments) continue;
             for (const seg of item.segments) {
                 if (seg.selected) {
-                    return true;
+                    const prev = seg.getPrevious();
+                    const next = seg.getNext();
+                    const isCurved =
+                        (!prev || seg.handleIn.length > 0) &&
+                        (!next || seg.handleOut.length > 0) &&
+                        (prev && next ? seg.handleOut.isColinear(seg.handleIn) : true);
+                    if (!isCurved) items.push(seg);
                 }
             }
         }
-        return false;
+        return items;
+    }
+    _getSelectedUnpointedPoints () {
+        const points = [];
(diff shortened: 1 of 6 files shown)


src/containers/mode-tools.jsx
```diff
@@ -13,10 +13,126 @@ class ModeTools extends React.Component {
...
+            const prev = point.getPrevious();
+            const next = point.getNext();
+            const noHandles = point.handleIn.length === 0 && point.handleOut.length === 0;
+            if (!prev && !next) {
+                continue;
+            } else if (prev && next && noHandles) {
+                // Handles are parallel to the line from prev to next
+                point.handleIn = prev.point.subtract(next.point)
+                    .normalize()
+                    .multiply(prev.getCurve().length / 2);
```
paulkaplan: Since the length of the resulting handles is largely arbitrary, I think we should use a value that has a neat geometric property: if you have a rectangle/circle, converting all the points to curved turns it into an oval/circle. 
Right now half the path length does this
handles-too-large

But we could make it do this
handles-just-right

Right now we are scaling the prev curve length by 0.5, but there is a magic number to get close to a circle: `0.3902628565`. 

_/warning math_
Arrive at that by looking at how to approximate a circle with bezier curves [source](http://spencermortensen.com/articles/bezier-circle/). A unit circle is approximated by:
```
P_0 = (0,1), P_1 = (c,1), P_2 = (1,c), P_3 = (1,0)
c = 0.551915024494
```
We are trying to turn a square (side length `s`) into a circle with the same anchors, which will be a larger circle with radius `s * sqrt(2) / 2`. So the correct scaling for the new handles is `c * sqrt(2) / 2`, or `0.3902628565`. Turns out this works with rectangles => ovals also, although I haven't done the math on why...

I checked with Carl, and FWIW this is actually the behavior from other vector editors like Sketch as well.

src/containers/mode-tools.jsx
```diff
@@ -13,10 +13,126 @@ class ModeTools extends React.Component {
...
+            } else if (prev && !next && point.handleIn.length === 0) {
+                // Point is end point
+                // Direction is average of normal at the point and direction to prev point, using the
+                // normal that points out from the convex side
+                // Lenth is curve length / 2
+                const convexity = prev.getCurve().getCurvatureAt(.1) < 0 ? -1 : 1;
```
paulkaplan: Should this use the `getCurvatureAtTime` function here, since we are using getNormalAtTime below? Is there a difference? The documentation is kind of confusing on whether getCurvatureAt takes a physical distance or a time parameter...

paulkaplan (commented):
A couple of questions. I think the smaller magic number scaling makes the handles a more manageable size in addition to a cool geometric property, but I'm not married to the idea if you don't like it.

src/containers/mode-tools.jsx
fsih: Sounds good! .5 always felt a bit too long.

src/containers/mode-tools.jsx
fsih: The documentation didn't mention getCurvatureAtTime, thanks! That makes more sense.

paulkaplan (dismissed):
LG

PR #243, opened 2017-12-21
Flip

fsih:
Fixes https://github.com/LLK/scratch-paint/issues/39

Implement flip

flipper


mewtaylor (changes requested):
This looks awesome! One small thing – the text looks like it's overflowing really close to the dashed border
screen shot 2017-12-27 at 9 15 12 am
I'm not sure what to do with this, since the buttons are equally spaced. I'm happy to leave it, not sure if we wanted to limit the width of the text though?

mewtaylor (dismissed):

fsih:
I wrapped flip horizontal/vertical in an input group, and it seems to look right now:
image

PR #244, opened 2017-12-29
Fix GH-226: set enterExitTransitionDurationMs to imperceptibly small value

mewtaylor:
#226 appears to be caused when the duration in ms is below a certain value, which is why we don’t see it in other popovers. I’m not sure exactly what may be causing this in the `react-popover` library and will open an issue about it, but for now, 20ms is imperceptibly fast, and appears to remove the error.

### DIFF
--- a/src/components/paint-editor/paint-editor.jsx
+++ b/src/components/paint-editor/paint-editor.jsx
@@ -246,7 +246,7 @@ const PaintEditorComponent = props => {
                             <InputGroup>
                                 <Dropdown
                                     className={styles.modUnselect}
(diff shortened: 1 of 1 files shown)


fsih (approved):
Just checking -- is there a reason we don't leave this as the default like in our other dropdowns? Should the other dropdowns match?

mewtaylor:
@fsih we could have it transition in like other `Popovers` – the main reason it's different here is that this isn't really a `Popover` like the other components, but meant to be using that library to implement a dropdown, which we didn't have transitions for. /cc @carljbowman

PR #251, opened 2018-01-11
Clean up the cursor preview when it is removed from its parent

fsih:
Fixes https://github.com/LLK/scratch-paint/issues/225


paulkaplan (approved):
LG

Issue #253, opened 2018-01-12
On an empty backdrop, the first thing you draw is not undoable


fsih:
1. Click on backdrop
2. Draw anything, say a brush stroke
3. Undo button doesn't turn on.
4. If you make more strokes, you can undo those, but undo gets disabled when you get back to the first stroke.

PR #254, opened 2018-01-12
Add an initial undo state when importing a blank costume

fsih:
Fixes https://github.com/LLK/scratch-paint/issues/253

The first undo state is considered the initial state and is not undoable. Add the initial snapshot for blank SVGs.


paulkaplan (approved):
LG

PR #263, opened 2018-01-21
Add holding shift to force horizontal scrolling

### LINKED ISSUE #156: Holding shift while scroll-wheeling should pan left/right (Windows Chrome)
Otherwise, it's impossible to scroll left and right with the scroll wheel. Several other paint editor programs use shift-scrolling as a way to scroll left and right.

(Is there an issue for this in scratch-blocks? Shift-scrolling should also be implemented there, especially if it is here.)

towerofnix:
Fixes #156.

This is a "workaround" for browsers which do not automatically set `deltaX` when the shift key is held (seen on Firefox-Linux and Chrome-Windows, but not Chrome-Mac). This PR makes it so that when the shift key is held while scrolling, `deltaX` is ignored, and the distance scrolled horizontally is set to `deltaY`.

This might have a strange effect on macbooks with trackpads (suddenly the distance you move your fingers left/right has no effect and horizontal scrolling is based on up/down finger movement), but I think that's acceptable - it would be obvious that it only behaves that way when shift is held, and wouldn't be too difficult to intuit anyways.

Tested locally on Firefox/Linux.

### DIFF
--- a/src/containers/paper-canvas.jsx
+++ b/src/containers/paper-canvas.jsx
@@ -164,7 +164,7 @@ class PaperCanvas extends React.Component {
                 new paper.Point(offsetX, offsetY)
             );
             zoomOnFixedPoint(-event.deltaY / 100, fixedPoint);
-        } else if (event.shiftKey) {
+        } else if (event.shiftKey && event.deltaX === 0) {
             // Scroll horizontally (based on vertical scroll delta)
             // This is needed as for some browser/system combinations which do not set deltaX.
             // See #156.
--- a/src/containers/paper-canvas.jsx
+++ b/src/containers/paper-canvas.jsx
@@ -164,6 +164,12 @@ class PaperCanvas extends React.Component {
                 new paper.Point(offsetX, offsetY)
             );
             zoomOnFixedPoint(-event.deltaY / 100, fixedPoint);
+        } else if (event.shiftKey) {
+            // Scroll horizontally (based on vertical scroll delta)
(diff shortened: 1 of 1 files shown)


fsih (changes requested):
I tried this out in Chrome on Windows, and it works well with mouse scroll!

On Windows and Mac, the touchpad behavior is as you described, where it only scrolls horizontally if you hold shift. I agree that that case seems less important.

On Mac though, this breaks horizontal mouse scroll, since holding shift causes scrolling to do nothing /:

towerofnix:
Thanks for testing, @fsih!

I made a change which only runs this workaround if event.deltaX is zero (and shift is pressed, of course). In that case, we know that the user probably wants to scroll horizontally, but deltaX isn't being set, so the "scroll horizontally by deltaX" block certainly won't work, and we have a go at using deltaY as the horizontal scroll distance.

towerofnix:
This does make trackpad-scrolling while shift is held just a little more funky, though, since now the workaround only applies if the user is scrolling strictly vertically. So when you're scrolling to the side at all, it goes back to using the normal scroll method, and the horizontal movement works as you'd normally expect. But if you're scrolling strictly *vertically* while shift is pressed, the editor pans horizontally!

fsih (approved):
Cool, looks good!

I tried it out on a Mac touchpad, and the scroll kind of jitters sometimes but works almost normally. Maybe if you're resizing something with shift held, and it's going off the edge, you might want to scroll while holding shift, but that's pretty out there.

PR #264, opened 2018-01-21
Use sans-serif font in playground

bates64:
Changes the playground body font to the Helvetica stack.

Before there were some cascade issues where you'd see default fonts in components (i.e. serif). This change is **only for the playground** as, I imagine, in scratch-3.0 this font style would be inherited from somewhere there.

Before (firefox-linux):
image

After:
image

### DIFF
--- a/src/playground/index.ejs
+++ b/src/playground/index.ejs
@@ -4,6 +4,7 @@
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <title><%= htmlWebpackPlugin.options.title %></title>
+    <style>body { font-family: "Helvetica Neue", Helvetica, Arial, sans-serif; }</style>
   </head>
   <body>
   </body>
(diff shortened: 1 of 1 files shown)


fsih (changes requested):
Thanks! Can you just format it by adding newlines after <style> and before </style> and indentation and spaces in the style of our .css files, such as https://github.com/LLK/scratch-gui/blob/0e56f866e5a4a736f7e618571fc64315a015fea3/src/index.css ?

thisandagain:
Closing due to lack of activity.

bates64:
@thisandagain oops, I forgot about this! Should have fixed the style now - if this PR is re-opened [the commit](https://github.com/heyitsmeuralex/scratch-paint/commit/ef78584045e00a883d8b37665ae50d19425fa837) should show up.

fsih (approved):
Looks good, thanks!

PR #265, opened 2018-01-22
Update the readme

fsih:
I think the current one might have been written back when the entire paint editor was just the word BANANAS


paulkaplan (approved):
👍

PR #266, opened 2018-01-23
Use translate instead of position for costume centering

fsih:
This fixes the centering of some sprites, most notably the refrigerator backdrop, but also some sprites where the center seemed to move around in paint but not on the stage. The issue was that item.position changes when the viewbox is removed, so we want to use translate instead.

before:
centering_before

after:
centering


cwillisf (approved):
So satisfying to see Casey dancing only in the ways the artist intended!
💃🕺💃🕺💃🕺

Issue #206, opened 2017-11-30
"Winter" backdrop is offset in editor (but not stage)


ericrosenbaum:
Expected: Backdrops are centered the same way in the editor and on the stage.

Actual: Add the "winter" backdrop (after [this PR](https://github.com/LLK/scratch-gui/pull/975)), and you see it display correctly on the stage but offset in the paint editor:

<img width="1276" alt="screen shot 2017-11-30 at 2 18 18 pm" src="https://user-images.githubusercontent.com/567844/33450419-bb318086-d5d9-11e7-8267-69b7ac8117d0.png">

thisandagain:
I think this is an issue with how some of the old SVG assets are reporting scale and rotation center. Some of them (like this one) go outside the bounding box defined for the SVG. I'll attempt a fix but it would be nice if the paint editor understood what to do in instances like this.

fsih:
I think it's that I'm doing the calculation for where the center is wrong. I had assumed that sprites were defining their centers relative to their own bounding boxes, but this sprite seems to be using somewhere else as its "origin". It's reporting its rotation center as (240, 180) or exactly the middle of the stage size. But its size is (616, 503), much bigger than the stage size. It's getting centered (240, 180) from its top left corner.

How do sprites define their rotation center?

thisandagain:
@fsih It is in the sprite / costume / backdrop definition here:
https://github.com/LLK/scratch-gui/blob/develop/src/lib/libraries/backdrops.json#L151

fsih:
I meant, what point is that point defined relative to? I guess it's not the upper left corner of the SVG.

thisandagain:
Assigning @fsih @paulkaplan and I for discussion. We should sit down and step through this together as it impacts multiple parts of the system (paint, render, libraries, asset creation).

fsih:
I noticed it's also offset in the library:
<img width="478" alt="screen shot 2017-12-28 at 12 19 29" src="https://user-images.githubusercontent.com/2855464/34417981-7f46f97e-ebc9-11e7-826f-c83355b481a0.png">

Issue #102, opened 2017-10-25
Costumes from Scratch 2.0 sometimes don't have rotation point imported to the same spot


fsih:
See sprite3 in project 181873537. The 2 costumes are reflected over the center in the paint editor on Scratch 2.0, but offset slightly to the left in the 3.0 paint editor

PR #267, opened 2018-01-25
Fix point rotation point is relative to, which is not viewBox upper left

fsih:
Fixes https://github.com/LLK/scratch-paint/issues/206
Fixes https://github.com/LLK/scratch-paint/issues/102

Fix point rotation point is relative to, which is not viewBox upper left. It's actually (0, 0) of where the viewbox itself is defined relative to.
ezgif-2-a827452094

### DIFF
--- a/src/containers/paper-canvas.jsx
+++ b/src/containers/paper-canvas.jsx
@@ -98,6 +98,20 @@ class PaperCanvas extends React.Component {
             svg = svg.replace(
                 '<svg ', '<svg xmlns="http://www.w3.org/2000/svg" ');
         }
+
+        // Get the origin which the viewBox is defined relative to. During import, Paper will translate
+        // the viewBox to start at (0, 0), and we need to translate it back for some costumes to render
+        // correctly.
+        const parser = new DOMParser();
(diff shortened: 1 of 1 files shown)


fsih:
@cwillisf we needed the viewbox numbers before they went through SvgRenderer

cwillisf (approved):
ROCKIN'!

Nitpick: the regex match / for-loop construct looks a bit fragile to my eye. Specifically, it looks like it handles well-formed SVG both with and without a viewbox, but what if the viewbox is malformed? Like, it's fewer than 4 numbers, or it's "0 oof 42 blah"?

If the answer is that none of this code deals with malformed SVG well and we should handle that as a separate issue (or several separate issues) then that's fine too :)

fsih:
Those 2 cases should get handled, by
`if (viewBox && viewBox.length >= 2 && !isNaN(viewBox[0]) && !isNaN(viewBox[1])) {`
In this case it ignores the viewbox, which is what I found Illustrator does with the SVGs.

I agree though that the manual parsing isn't great and our SVG edge case handling here is getting out of hand. I'll work on adding tests for import. I made https://github.com/LLK/scratch-paint/issues/275

Issue #247, opened 2018-01-05
Flip and copy buttons should act on everything when nothing is selected


ericrosenbaum:
Currently, when nothing is selected, the flip buttons and the copy button are disabled.

Instead, they could be always enabled. If nothing is selected, the flip buttons would flip everything. The copy button would copy everything.

PR #268, opened 2018-01-25
Flip all

fsih:
Fixes https://github.com/LLK/scratch-paint/issues/247

When nothing is selected, flip everything in the paint editor over the rotation point.
flipall

### DIFF
--- a/src/components/mode-tools/mode-tools.jsx
+++ b/src/components/mode-tools/mode-tools.jsx
@@ -151,13 +151,11 @@ const ModeToolsComponent = props => {
                 </InputGroup>
                 <InputGroup className={classNames(styles.modLabeledIconHeight)}>
                     <LabeledIconButton
-                        disabled={!props.selectedItems.length}
                         imgSrc={flipHorizontalIcon}
                         title={props.intl.formatMessage(messages.flipHorizontal)}
                         onClick={props.onFlipHorizontal}
                     />
                     <LabeledIconButton
-                        disabled={!props.selectedItems.length}
                         imgSrc={flipVerticalIcon}
                         title={props.intl.formatMessage(messages.flipVertical)}
                         onClick={props.onFlipVertical}
--- a/src/components/mode-tools/mode-tools.jsx
+++ b/src/components/mode-tools/mode-tools.jsx
@@ -151,13 +151,11 @@ const ModeToolsComponent = props => {
                 </InputGroup>
                 <InputGroup className={classNames(styles.modLabeledIconHeight)}>
                     <LabeledIconButton
-                        disabled={!props.hasItems}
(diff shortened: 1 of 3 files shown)


ericrosenbaum:
My intuition about this is that it should work the same as when everything is selected, and flip around the bounding box center, not around the rotation point.

What do you think @carljbowman?

fsih:
One argument against that is how the refrigerator background would behave. This is a bit of an extreme example because the fridge background has a lot of extra stuff sticking out of the sides, but many of the backgrounds would be just a bit off.
fridgeflip

I think if the rotation point is set intentionally, flipping around the rotation point is more likely to seem right. I guess the question is how many people won't understand the rotation point.

carljbowman:
Even seeing the backdrop example, I agree with @ericrosenbaum that the initial experience should be that the artwork flips relative to the selected rotation point. Flipping a costume feels like the more frequent use case.

Although this does raise the question around backdrops larger than 480x360? We could explicitly crop all backdrops to be 480x360. Seems the abnormal backdrops (e.g. "blue sky") are causing lots of problems.

thisandagain:
Agreed with both @carljbowman and @ericrosenbaum re: staying relative to the selected rotation point.

+1 to cropping the existing backdrops. I can add that to my to-do list.

fsih:
Even if we crop our built-in backgrounds, I feel like when you draw a background it's natural to have items extend off the edges, the way that Party has balloons extending off the top edge and Winter has trees extending off the sides. User-generated or user-edited backgrounds would still have this issue. Would we want to cut off those edges as the user is editing in the paint editor?

fsih:
(I'm reverting the change where the buttons are disabled when there's nothing in the paint editor, because there's a bug where the paint editor isn't listening on hasSelectableItems state and disabled state doesn't always get updated, and I don't want to make that big a change in this PR)

ericrosenbaum:
I think our conversation became a bit confused because we are using words differently- here's an attempt to clarify (not sure if these are the best terms to use):

**Rotation center** - this is the gray crosshair in the center of the grid, representing the center point for rotation of the sprite when it is on the stage. 

**Bounding box center** - This is the center of the rectangle formed by selected artwork in the paint editor.

Generally, when you select something and flip it, it flips around the bounding box center.

In this PR, when nothing is selected, it flips around the rotation center.

I was suggesting that the flip buttons should always flip around the bounding box center. When nothing is selected, it should compute this bounding box as though everything is selected.

fsih:
Okay I've changed it to not rotate around the rotation center

fsih:
@ericrosenbaum is this good now?

ericrosenbaum (approved):
!Looks Good To MeeM oT dooG skooL!

Issue #82, opened 2017-10-20
Holding shift should draw straight or 45 degree lines in line tool


fsih:

kyleplo:
This doesn't seem to work anymore.

PR #269, opened 2018-01-25
Hold shift to lock line tool to 45 degree angles

fsih:
Fixes https://github.com/LLK/scratch-paint/issues/82

lines45

### DIFF
--- a/src/containers/line-mode.jsx
+++ b/src/containers/line-mode.jsx
@@ -13,6 +13,7 @@ import {changeStrokeWidth} from '../reducers/stroke-width';
 import {changeMode} from '../reducers/modes';
 import {clearSelectedItems} from '../reducers/selected-items';
 import {MIXED} from '../helper/style-path';
+import {snapDeltaToAngle} from '../helper/math';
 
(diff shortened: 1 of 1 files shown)


paulkaplan:
Code looks good, and feels good to use. Nice catch on having straight lines take priority over distance snapping, that seems right.

One question: would we want this to also work for the brush tool? Or no? /cc @carljbowman

carljbowman:
@paulkaplan - Good question. Holding shift with more free form tools, like the paint brush, is not as much a convention. I lean toward not including, because I think it would take some time to dial in the UX. 

The line tool has a clear origin point that becomes fixed when you start to draw a path. That point can act as a pivot when drawing a shape. I am struggling to think of how that model would be conveyed in the brush tool. Especially if you have already started to draw and then hold shift.

In the case of the line tool, if you start drawing and then hold shift, the line snaps to a 45 degree.

In the case of a shape tool, if you start drawing and then hold shift, the shape retains the ratio of the drawn shape. Or if you hold shift and begin drawing as shape, the convention is to retain a square ratio.

Issue #236, opened 2017-12-19
Nudge


fsih:
Would like to be able to move things in select states very small amounts using the arrow keys.

PR #271, opened 2018-01-25
Nudge

fsih:
Fixes https://github.com/LLK/scratch-paint/issues/236
- getSelectedSegments is broken out since it's shared by point tool and reshape tool
- nudge-tool is broken out since it's shared by select-tool, oval-tool and rect-tool

Use arrow keys in select modes to nudge
nudge

### DIFF
--- a/src/helper/selection-tools/nudge-tool.js
+++ b/src/helper/selection-tools/nudge-tool.js
@@ -1,0 +1,50 @@
+import {getSelectedRootItems} from '../selection';
+import paper from '@scratch/paper';
+
+/**
+ * Tool containing handlers for arrow key events for nudging the selection.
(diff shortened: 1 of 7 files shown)


paulkaplan (approved):
Looks good. Did you want to add something like "shift to supernudge"? @carljbowman

carljbowman:
This is amazing!!! So excited.

I leave it up to @fsih to decide on the super nudge. Nice to have, but not necessary as I know there are other things to work on.

fsih:
It sounds like it's not hard to add, but what's super nudge? How much more should it nudge?

paulkaplan:
We can do it in a follow-up, sorry i should have asked this IRL instead of derailing the review.

LGTM!

Issue #276, opened 2018-01-29
Using eyedropper triggers mini-editor


ericrosenbaum:
Clicking on the color picker's eyedropper tool reveals mini-paint-editor mode.

- Go to https://llk.github.io/scratch-gui/develop/ and select costumes tab.
- Select fill menu, select eye dropper, select a color.
- Click again to dismiss fill menu.
- Mini-editor!

![mini-paint-bug2](https://user-images.githubusercontent.com/567844/35537979-46b72a2e-051a-11e8-876d-a84d7b6437e4.gif)

(forgot to add platform details when I posted this, sorry!)
Mac OS 10.13.2 Chrome 64.0, retina display, external monitor NOT connected

paulkaplan:
Yikes. Confirm I can repro, also on preview.

fsih:
Interesting. I haven't been able to repro at all on Windows (Chrome 63.0.3239.132/Firefox/Edge)

paulkaplan:
I repro'd on mac/chrome, but couldn't repro with mac/safari, so could be a browser/platform thing?

fsih:
I tried on mac/chrome and couldn't repro, develop or preview, Mac 10.12.6  and Chrome 63.0.3239.84 (Also Firefox/Safari)

ericrosenbaum:
I just tested this again and noticed that it only happens when NOT using an external monitor (i.e. when using the laptop's retina display)

paulkaplan:
@ericrosenbaum Oh dear, that sounds like it could be a deep graphics card level issue. Kind of like how we used to see the green flag get patterned across the block workspace. @fsih fwiw I'm on chrome 64. Could be a bug introduced on Chrome 64, which would explain why we haven't seen this previously (seems like it would have come up. chrome 64 came out last week).

fsih:
I reproed it! By upgrading to Chrome 64.0.3282.119 (Mac)

fsih:
I found a bug in your bug report! See lower right corner
![flashing](https://user-images.githubusercontent.com/2855464/35593219-488cdd62-05dd-11e8-93f3-d607fa9ad3d4.gif)

fsih:
Filed https://bugs.chromium.org/p/chromium/issues/detail?id=807427 on Chrome

fsih:
Fixed in Chrome 65

PR #283, opened 2018-02-07
Temporary fix for eyedropper issue in Chrome 64

paulkaplan:
Buffer the paper canvas offscreen for picking to ensure safe access.

This gets around the bug #276 where chrome cannot handle getImageData
off a canvas on HDPI screens. In the interest of having a single
code-path, and because does not impose a large performance burden, I
think we should just use it for all platforms until chrome is really
fixed.

Also it is important to note that because this bug has to do with how
they handle accelerated 2d canvas contexts, it does not impact the stage
color picking, so that does not need to be changed.

The early returns I added are because the `getColorInfo` now potentially
returns null if the buffered image has not been loaded yet.

### DIFF
--- a/src/containers/paint-editor.jsx
+++ b/src/containers/paint-editor.jsx
@@ -176,6 +176,7 @@ class PaintEditor extends React.Component {
                 this.eyeDropper.pickY,
                 this.eyeDropper.hideLoupe
             );
+            if (!colorInfo) return;
             if (
                 this.state.colorInfo === null ||
                 this.state.colorInfo.x !== colorInfo.x ||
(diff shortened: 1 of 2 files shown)


fsih:
By the way, on that Chrome bug, they merged the fix to M64, and they're going to do a hotfix push next week!

paulkaplan:
@fsih that is great news. Do you think we should merge this and roll it back out when the fix is out?

fsih (approved):
This works really well, thanks Paul! It might not work when we have the dynamically sized paper canvas working, which is supposed to come in with the scrolling update in March, but it should be safe to roll back by then.

fsih (changes requested):
Oops, my comment made me realize that you can already do that, sorry. This won't work if you are zoomed in and scroll the canvas before eyedropping.

fsih:
That use case is already pretty broken as I found in https://github.com/LLK/scratch-paint/issues/288 ... so maybe it doesn't matter much.

fsih (approved):
@paulkaplan and I decided it's ok given that it's already pretty broken if you scroll after eyedropping, and it's an edge case.

Issue #278, opened 2018-02-05
Bounding box becomes fixed when holding Shift


carljbowman:
## Current Behavior
When dragging artwork in the paint editor the bounding box stays with the artwork, but when holding shift the bounding box becomes fixed and the handles remain.

### Dragging artwork
**Just dragging artwork**
![dragging](https://user-images.githubusercontent.com/3409578/35819441-b9f963ac-0a70-11e8-811a-2eb2815127a6.gif)

**Dragging artwork and then holding shift**
![dragging shift](https://user-images.githubusercontent.com/3409578/35819512-f497454c-0a70-11e8-8256-fe806f384fb1.gif)

### Rotating
**Just rotating artwork**
![rotating](https://user-images.githubusercontent.com/3409578/35819714-870ddb34-0a71-11e8-9b34-c85c3a26f469.gif)

**Rotating artwork and then holding shift**
It also feels a bit awkward that you rotate an element slightly and then hold shift, the fixing to the 45° angles become relative to the current rotation. I'd expect the 45° fixing would be relative to the original artworks rotation. 

![rotating shift](https://user-images.githubusercontent.com/3409578/35819765-a775037a-0a71-11e8-9ca4-06df7c283d43.gif)

PR #284, opened 2018-02-07
Fix bounding box issue when using shift+drag

paulkaplan:
Only update the selection bounds if a nudge actually occurs

This was causing the problem where holding shift while dragging would
trigger the selection box to be shown, interrupting the drag.

Fixes the dragging part of https://github.com/LLK/scratch-paint/issues/278, but not the rotation part.


fsih (approved):
This looks like a good fix for now. I've logged https://github.com/LLK/scratch-paint/issues/286 to fix it better.

Issue #282, opened 2018-02-05
Clicking into the Workspace does not clear focus from inputs


carljbowman:
**Repro** 
1. Click into an input
2. Select something or click in the workspace
3. Field retains focus

Can lead to weird issues like this:
![input focus](https://user-images.githubusercontent.com/3409578/35820811-d242b9aa-0a74-11e8-98bb-924ae8487a2e.gif)
1. Click into the input
2. Select artwork
3. Field retains focus
4. Hit the delete key to delete artwork, but the outline size is deleted

PR #285, opened 2018-02-07
Don't nudge selection when input is focused

### LINKED ISSUE #280: Incrementing the outline size with arrow keys also nudges the artwork
**Repro**
1. Select an object
2. Click into the outline size
3. Use up arrow to increment the size

nudge increment

paulkaplan:
Fixes https://github.com/LLK/scratch-paint/issues/280

@carljbowman @fsih  I think we should talk in person about whether it is ok to bring this in without also doing https://github.com/LLK/scratch-paint/issues/282, which might cause a lot of frustration. The opposite might be a better short-term fix (make the input a "text" type so that it doesn't respond to keyboard controls, allowing the nudge tool to work despite the focus). The reason for that is I think #282 is going to be much more complicated to solve. 

@fsih I'll tag this needs-discussion until we actually talk about it, so no need to review it before then.

### DIFF
--- a/src/helper/selection-tools/nudge-tool.js
+++ b/src/helper/selection-tools/nudge-tool.js
@@ -15,8 +15,10 @@ class NudgeTool {
         this.onUpdateSvg = onUpdateSvg;
     }
     onKeyDown (event) {
-        // Bail if the real event target is an input
-        if (event.event.target.tagName === 'INPUT') return;
+        if (event.event.target instanceof HTMLInputElement) {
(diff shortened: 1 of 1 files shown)


fsih (dismissed):
Maybe we should change the syntax here to match the other place in scratch-paint,
if (event.target instanceof HTMLInputElement) {
but I don't feel strongly about it

paulkaplan:
@fsih fixed.

fsih (approved):

PR #289, opened 2018-02-08
Change pan/zoom bounds to 360/480 to match the real bounds.

### LINKED ISSUE #223: At 100% zoom, you can scroll up in the paint editor
image

paulkaplan:
Fixes https://github.com/LLK/scratch-paint/issues/223 a bug where you can scroll the paint editor outside the 480/360 bounds because it allows you to go to 500/400. 

I know this will be fixed with the other scrollbar work, it just was a thorn in my side that you could still do this, and it is such a simple fix 💯 

You can no longer do this:
image

### DIFF
--- a/src/helper/view.js
+++ b/src/helper/view.js
@@ -9,11 +9,11 @@ const clampViewBounds = () => {
     if (top < 0) {
         paper.project.view.scrollBy(new paper.Point(0, -top));
     }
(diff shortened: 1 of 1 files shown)


fsih (approved):
Great! Thanks Paul! That work was probably not going to happen for another month

PR #290, opened 2018-02-09
Mark all clickable images as not draggable.

### LINKED ISSUE #1407: Paint Editor Slider Gets stuck to Mouse Pointer
### Expected Behavior

Color picker slider should not get stuck to mouse pointer, it should only drag if the user is actively clicking down.

### Actual Behavior

If the slider (and text in the color picker) is selected, and the user drags in this selected state outside of the color picker, the slider gets stuck to the mouse pointer.

Needs a consistent repro, but see gif below.

### Steps to Reproduce

slider-weirdness

Mac OS 10.13.3 Safari 11.0.3

paulkaplan:
Resolves the costume tab part of https://github.com/LLK/scratch-gui/issues/1430
Also resolves https://github.com/LLK/scratch-gui/issues/1407

When using images as clickable elements (within buttons or alone) you
need to mark them as not draggable. They are draggable by default and
starting a drag prevents the onClick from being handled, messing up the
clickable behavior.

no-drag-images-paint

### DIFF
--- a/src/components/color-button/color-button.jsx
+++ b/src/components/color-button/color-button.jsx
@@ -29,11 +29,13 @@ const ColorButtonComponent = props => (
             {props.color === null ? (
                 <img
                     className={styles.swatchIcon}
+                    draggable={false}
                     src={noFillIcon}
                 />
             ) : ((props.color === MIXED ? (
                 <img
                     className={styles.swatchIcon}
+                    draggable={false}
                     src={mixedFillIcon}
                 />
             ) : null))}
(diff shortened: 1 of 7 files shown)


carljbowman (commented):
Great! Thanks for doing that so quickly!

carljbowman (approved):
Looks good to me.

fsih (commented):
Should images ever be draggable? Maybe we should add something higher level to prevent it, along with selectable?

paulkaplan:
@fsih maybe not, but there isn't a really effective way to make that happen. We could make a component that just wraps `img` and sets draggable to false. We should do that if/when (a) we want to do something with CSS in addition to this dom property or (b) we combine gui/paint components so we have a shared library of components, don't want to do that in both.

Issue #90, opened 2017-10-23
Set an undo stack size limit


fsih:

thisandagain:
If we have learned anything over the past 10 years of working on Scratch it is that if you don't set a limit then someone will do something crazy and either crash Scratch (when this stuff is server-side) or get very sad and crash their client. Yes, we should set a maximum undo stack size. 😉 

How would you feel about 99 as a limit?

thisandagain:
For reference, the default size of the undo stack in Adobe Photoshop is 50.

PR #300, opened 2018-02-27
Add an undo stack size limit to paint

fsih:
Fixes https://github.com/LLK/scratch-paint/issues/90
Limit the undo stack size to 100. That means you can undo up to 99 times.


paulkaplan (approved):
LG

Issue #23, opened 2017-09-26
Improve brush tool performance


fsih:
Note: keep this scoped to some easy fixes that help elevate the immediate issue.

Kenny2github:
I'll transfer info from my dupe issue:
Dragging the brush gets laggier and laggier, with the brush jumping around more and more.
Also from @towerofnix:
>Note this isn't a Raspberry Pi! I'm probably on a higher end computer, so, the problem might not be linked to available memory/CPU.

PR #304, opened 2018-03-01
Remove all view.draw and view.update

fsih:
View.draw and view.update can slow performance when called too often, and we seem to call it too often. In fact, I can't tell that anything changes at all when I take them all out. So let's try taking them all out.

This work is related to https://github.com/LLK/scratch-paint/issues/23


rschamp (approved):
It works!

PR #305, opened 2018-03-01
Brush efficiency

### LINKED ISSUE #23: Improve brush tool performance

fsih:
Resolves https://github.com/LLK/scratch-paint/issues/23

This PR improves the efficiency and shortcomings of the broad brush tool, and then increases the threshold for using the broad brush (Previously sizes below 9 would use the broad brush, and now it's sizes below 30)

### Efficiency
**tl;dr don't waste effort and do things in small chunks instead of all at the end**

The main problem here was that each time the path drawn became longer, the entire path would be re-smoothed. (paper.Path.smooth removes all handles from a path and recalculates them to generate a smooth line.) This meant we wasted more and more time smoothing, and most of it was repeated effort. Then at the very end, the entire path was simplified (paper.Path.simplify generates a path that passes close to all points on the existing path, while using as few control points as possible).

The improvement was to amortize path simplification. Every 20 steps, only whatever has been newly added to the path is simplified. We avoid wasted effort and a large simplify at the end. (Smoothing is unnecessary: since simplify ignores handles, running smooth before simplify has no effect)

The funny thing is, now that it's possible to draw more efficiently, it's easier to draw a massive scribble which will require a huge merge at the end that can freeze the UI for quite awhile (mostly from paper.Path.resolveCrossings and the large number of unites). We could improve on this in the future by amortizing these operations as well.

### Mitigating shortcomings
**tl;dr broad brush tries to draw rounded corners now so it looks ok at larger sizes, which is good because it's more efficient than the other brush**

The brush tool uses 2 brushes: the broad brush and the segment brush. The broad brush is more efficient, as it adds points to the path as a user moves the mouse, while the segment brush has to perform a merge on every mouse move. However, the broad brush doesn't deal with sharp turns well, which becomes really obvious at larger brush sizes, which is why we switch to the segment brush. (Trying to draw a small circle can easily result in a crazy star shape as the brush can't figure out the stroke direction.) It also leaves more work to the end, since it can cross its own path, and all that has to be unioned at the end.

I've improved on the appearance of flat edges by trying to detect sharp turns, and stamping a circle there whenever we make one. I was able to take advantage of amortized simplification work to avoid smoothing of those sharp corners, which used to cause those weird end blobs. This does make more work at the end, since all those circles need to be merged into the path, but it makes the appearance of the path good enough to draw at larger sizes.

### Really unscientific benchmarks
Before: scribble at size 5 for 10 seconds. Takes a couple seconds to resolve the merge. 2173 mouse moves handled.
scribblesize5preview
After: scribble at size 5 for 10 seconds. Takes like 10 seconds to resolve the merge ): 5234 mouse moves handled.
scribblesize5

Before: scribble at size 25 for 10 seconds. Resolves immediately. 832 mouse moves handled.
scribblesize25preview
After: scribble at size 25 for 10 seconds. Takes a couple seconds to resolve the merge. 1312 mouse moves handled.
scribblesize25

### DIFF
--- a/src/helper/blob-tools/blob.js
+++ b/src/helper/blob-tools/blob.js
@@ -428,8 +428,10 @@ class Blobbiness {
     }
 
     deactivateTool () {
-        this.cursorPreview.remove();
-        this.cursorPreview = null;
+        if (this.cursorPreview) {
+            this.cursorPreview.remove();
+            this.cursorPreview = null;
+        }
         this.tool.remove();
         this.tool = null;
     }
--- a/src/helper/blob-tools/blob.js
+++ b/src/helper/blob-tools/blob.js
@@ -94,7 +94,9 @@ class Blobbiness {
         };
         
         this.tool.onMouseDown = function (event) {
-            blob.resizeCursorIfNeeded(event.point);
+            blob.cursorPreview.remove();
+            blob.cursorPreview = null;
+            //blob.resizeCursorIfNeeded(event.point);
             if (event.event.button > 0) return; // only first mouse button
             this.active = true;
 
@@ -105,13 +107,13 @@ class Blobbiness {
                 blob.brush = Blobbiness.SEGMENT;
                 blob.segmentBrushHelper.onSegmentMouseDown(event, blob.tool, blob.options);
             }
-            blob.cursorPreview.bringToFront();
-            blob.cursorPreview.position = event.point;
-            paper.view.draw();
+            // blob.cursorPreview.bringToFront();
+            // blob.cursorPreview.position = event.point;
+            // paper.view.draw();
         };
 
         this.tool.onMouseDrag = function (event) {
-            blob.resizeCursorIfNeeded(event.point);
+            //blob.resizeCursorIfNeeded(event.point);
             if (event.event.button > 0 || !this.active) return; // only first mouse button
             if (blob.brush === Blobbiness.BROAD) {
                 blob.broadBrushHelper.onBroadMouseDrag(event, blob.tool, blob.options);
@@ -121,9 +123,9 @@ class Blobbiness {
                 log.warn(`Brush type does not exist: ${blob.brush}`);
             }
 
-            blob.cursorPreview.bringToFront();
-            blob.cursorPreview.position = event.point;
-            paper.view.draw();
+            // blob.cursorPreview.bringToFront();
+            // blob.cursorPreview.position = event.point;
+            // paper.view.draw();
         };
 
         this.tool.onMouseUp = function (event) {
--- a/src/helper/blob-tools/blob.js
+++ b/src/helper/blob-tools/blob.js
@@ -23,7 +23,7 @@ class Blobbiness {
     // Segment brush has performance issues at low threshold, but broad brush has weird corners
     // which get more obvious the bigger it is
     static get THRESHOLD () {
-        return 9;
+        return 30 / paper.view.zoom;
     }
 
     /**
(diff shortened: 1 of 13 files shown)


paulkaplan:
@fsih I am seeing a bit of strange behavior when you draw a quick stroke. Here is a gif for posterity, i can give you a demo IRL. Left side is this new version.

brush-performance

src/helper/blob-tools/blob.js
```diff
@@ -142,11 +141,9 @@ class Blobbiness {
                 blob.mergeBrush(lastPath);
             }
 
-            blob.cursorPreview.visible = false;
+            blob.cursorPreview.remove();
+            blob.cursorPreview = null;
             blob.onUpdateSvg();
-            blob.cursorPreview.visible = true;
-            blob.cursorPreview.bringToFront();
-            blob.cursorPreview.position = event.point;
```
paulkaplan, line 150: Is this commit related to the brush performance? Or is it a separate bug? If it is a separate bug, could you file a PR separately so that reviewing can be more linear? If it is a bug that came from something else you introduced here, it would be very helpful for reviewing to squash this commit with the commit that introduced the issue, or to otherwise mention in the commit message what this is addressing? It is hard to know exactly what is going on with this change.

src/helper/blob-tools/broad-brush-helper.js
```diff
@@ -171,23 +190,48 @@ class BroadBrushHelper {
...
+                // If the union of the two shapes is smaller than the original shape,
+                // or it caused the path to become a compound path,
+                // then there must have been a glitch with paperjs's unite function.
+                // In this case, skip merging that segment. It's not great, but it's
+                // better than losing the whole path for instance. (Unfortunately, this
+                // happens reasonably often to scribbles, and this code doesn't catch
+                // all of the failures.)
+                this.finalPath.insertAbove(temp);
+                temp.remove();
+                log.warn('Skipping a merge.');
```
paulkaplan, line 239: Is this log desired in production? If it was just for testing, can it be removed?

src/helper/blob-tools/broad-brush-helper.js
```diff
@@ -102,7 +121,7 @@ class BroadBrushHelper {
         const newPoints = Math.floor((length - this.smoothed) / 2) + 1;
 
         // Where to cut. Don't go past the rounded start of the line (so there's always a tempPathMid)
-        const firstCutoff = Math.min(newPoints + 1, Math.floor((length / 2) - 1));
+        const firstCutoff = Math.min(newPoints + 1, Math.floor((length / 2)));
```
paulkaplan: Can you explain this change?

paulkaplan (commented):
I had a few comments, and also a question about the end cap behavior that I posted with an attached gif.

rschamp (approved):
In my unscientific testing, I concluded this doesn't break anything, but I think a second review would be appropriate.  (Edit: ha, didn't see @paulkaplan's review come in while I had this PR open) The code makes sense to me.

src/helper/blob-tools/blob.js
fsih: Sorry. This was a bug I noticed while working on performance, but isn't very related. I've pulled it out into https://github.com/LLK/scratch-paint/pull/316

src/helper/blob-tools/broad-brush-helper.js
fsih: I thought it made sense to leave in for production. Unfortunately, it happens a lot. This will log when a union fails in paper.js (it doesn't catch all the failures though), which means that there is a bug that needs fixing somewhere. Right now those bad unions fail silently, and may be why when you draw a scribble the brush stroke suddenly disappears or inverts. Hopefully we can eventually work on debugging those bad unions in paperjs.

paulkaplan (approved):
LGTM

PR #323, opened 2018-03-14
Move the SVG in the playground back to the center of the editor

fsih:
It became offscreen when the definition of rotation center changed.

Before
image

After
image


kchadha (approved):
LG!

PR #324, opened 2018-03-14
Reduce the overlap to .1

fsih:
When you fill in a hole, the fill actually is made a tiny bit bigger than the hole, because otherwise you sometimes get tiny white gaps between the fill and the outline.

The goal of the overlap is
1. Reduce the amount of gaps visible between the fill and the outline
2. Don't increase the overlap so much that it's obviously overlapping

(note that the expand algorithm is pretty naive and won't be perfect if you draw a crazy spiky outline, it's just supposed to help with most cases)

In playtesting @carljbowman found that the overlap is too obvious when using width 1 brush strokes and filling them in. The overlap used to be .5, and this reduces it to .1 canvas units.

### DIFF
--- a/src/helper/math.js
+++ b/src/helper/math.js
@@ -82,17 +82,17 @@ const sortItemsByZIndex = function (a, b) {
     return null;
 };
 
-// Expand the size of the path by approx one pixel all around
-const expandByOne = function (path) {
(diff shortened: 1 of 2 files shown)


kchadha (approved):
LG!

PR #325, opened 2018-03-15
Unfocus text fields when canvas clicked and don't take keystrokes in canvas when text fields are focused

fsih:
Fix some issues where you can edit text and the paint editor at the same time.

### DIFF
--- a/src/containers/paint-editor.jsx
+++ b/src/containers/paint-editor.jsx
@@ -50,6 +50,10 @@ class PaintEditor extends React.Component {
     }
     componentDidMount () {
         document.addEventListener('keydown', this.props.onKeyPress);
+        // document listeners used to detect if a mouse is down outside of the
+        // canvas, and should therefore stop the eye dropper
+        document.addEventListener('mousedown', this.onMouseDown);
+        document.addEventListener('touchstart', this.onMouseDown);
     }
     componentDidUpdate (prevProps) {
         if (this.props.isEyeDropping && !prevProps.isEyeDropping) {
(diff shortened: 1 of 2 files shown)


src/helper/selection-tools/reshape-tool.js
```diff
@@ -232,6 +232,11 @@ class ReshapeTool extends paper.Tool {
         this.active = false;
     }
     handleKeyDown (event) {
+        if (event.event.target instanceof HTMLInputElement) {
+            // Ignore nudge if a text input field is focused
+            return;
+        }
```
rschamp, line 238: Is this here intentionally? Looks unrelated. Probably fine to leave it though.

src/containers/paint-editor.jsx
```diff
@@ -135,6 +141,10 @@ class PaintEditor extends React.Component {
         this.canvas = canvas;
     }
     onMouseDown () {
+        if (document.activeElement instanceof HTMLInputElement) {
```
rschamp: To keep this contained to this specific issue, would you also check that the mousedown event is on the paint editor canvas?

rschamp (changes requested):

src/helper/selection-tools/reshape-tool.js
fsih: This is vaguely related. Currently, if you are on the reshape tool and you have something selected, and you also have a textbox focused, if you press the arrow keys, both the selected thing and the cursor position will move. This return makes it so that only the cursor position moves. I will update the description.

src/containers/paint-editor.jsx
fsih: Oops, thanks for catching

rschamp (approved):

Issue #25, opened 2017-09-26
Add text tool


fsih:

Kenny2github:
Suggestion: Allow Unicode characters! Currently, for example, to type Chinese text I have to use an external image with *rasterized Chinese characters* and import it. If we could type Unicode characters this would help massively for i18n of Scratch projects.

carljbowman:
## Scratch 2.0
### Functionality
Below is a high-lever overview of the 2.0 text tool's functionally and how text interacts with other tools.

**Text Tool** 

- Primary Tooling
  - Add/Edit text directly in the workspace
  - Uses conventional text editing UI (e.g. highlighting, cursor…)
- Mode Tools
  - Font menu to set font
  - Color picker to set/remove the color for all the text
    - Cannot highlight individual characters and set color
- Key commands
  - Copy / Pasta  (OS hotkeys)
  - Delete (delete key)
- MISC Notes / States
  - Single space text objects cannot exist. More than one space or new lines (returns) can exist
    1. Click to make a new text object
    2. Choose the Select Tool
    3. Drag select the are when you made the text object — No text object

**Select Tool** 
- When text is selected and using bounding box, one can stretch/rotate text in the workspace like any other vector object
  -  The text does not become an expanded shape or rasterized as a result of stretching/rotating. Instead it remains editable text and retains stretch/rotation when editing the text.

**Fill Tool**
- When using the Fill tool, all of the text fills the selected color
  - Cannot Fill individual characters 

## Scratch 3.0 - Phase 1
### Functionality
Below is a high-lever overview of the 3.0 text tool's functionally and how text interacts with other tools. Sketches below to help contextualize. 

**Text Tool**
- Primary Tooling
  - Add/Edit text directly in the workspace
    - Uses conventional text editing UI (e.g. highlighting, cursor…)
    - What happens when you click out of “Text Editing State”?
      - 2.0 has inconsistent behavior: sometimes it creates a new text box, something it just deselects the text entirely, and sometimes it transitions to a “Selected State”
      - 3.0 should transition to a “Selected State” with Text mode tools — **[ Different then 2.0 ]** 
- Mode Tools
  - Font menu to set font
  - Set text size — **[ Different then 2.0 ]** 
  - Color picker to set/remove the color for all the text
    - Cannot highlight individual characters and set color
- Key commands
  - Copy / Pasta  (OS hotkeys) — **There may be dragons here**
  - Delete (delete key)
- MISC Notes / States
[...]

PR #333, opened 2018-03-22
Text edit

fsih:
Progress toward https://github.com/LLK/scratch-paint/issues/25
Adds text editing to the text tool!

See diff from initial text tool PR here:
https://github.com/fsih/scratch-paint/compare/textTool...fsih:textEdit?expand=1

When text is in edit mode, add a text area directly on top of it. This way we get all text editing capabilities of text areas.

Some notes
- The text edit area needs to stay on top of the text appearing on the canvas, no matter what orientation it's in. If the text area is able to be scrolled or wraps, it will be out of sync. The text can also have some crazy transformations applied to it through the select tool. This math is tricky, so this should be thoroughly tested.
- Zooming in and scrolling around the canvas also changes the matrix of the text area. I needed to add state to keep track of when that happened (state.scratchPaint.viewBounds)
- Added undo state to the text tool. When the user stops typing for 1 second, a new undo state is created. This is a pretty long timeout. I made it this long because sometimes kids aren't familiar with keyboard layouts. We might want to play with this number.


fsih:
I made the text tool draw no outline by default, as recommended by Paul

paulkaplan:
Just a few things I noticed playing with this. I'll probably create separate issues after this is merged, but wanted to have them down here for posterity. 

When clicking another text area while editing a text area, the currently editing text area should both shift out of editing mode (which it does) _and_ shift out of select mode (which it does not).
text-select-1

When in text editing mode, after exiting a text area (so that it is selected but not editing), you cannot re-enter editing on that text area without clicking outside to deselect it first
text-select-2

When in the normal select mode, there is a heavy expectation that double clicking on text would enter editing mode (from most GUIs with text areas, including scratch 2), but it doesn't. 
text-select-3

When in normal select mode, with a text area selected, clicking the text mode tool in scratch 2 transitions that text area into editing mode. Here it simply deselects everything. In the case where 
 multiple things are selected, scratch 2 deselects everything. 
text-select-4

paulkaplan:
Fixed the first and second issue with a commit: af2c49e

src/helper/tools/text-tool.js
```diff
@@ -0,0 +1,326 @@
...
+    }
+    handleMouseMove (event) {
+        const hitResults = paper.project.hitTestAll(event.point, this.getTextEditHitOptions());
+        if (hitResults.length) {
+            document.body.style.cursor = 'text';
+        } else {
+            document.body.style.cursor = 'auto';
+        }
+    }
+    handleMouseDown (event) {
```
paulkaplan, line 138: There is some seriously complicated logic in this function. I'd like for us to come back to this and break it into more clear state transitions. It took me ~20 minutes to understand the entire FROM => TO transitions!

paulkaplan (approved):
This LG for now. There are obviously some follow ups we'll need to make (a few were listed in the comments) but this feels good enough to start testing with.

PR #335, opened 2018-03-26
Redraw select bounding box controls on zooming #219

### LINKED ISSUE #219: Select bounding box scales with zoom
**Currently on Develop**
Browser: Chrome 63.0.3239.84
OS: Mac OS X 10.11.6
bounding box sizing with zoom

1. Select an object
2. Zoom in or out
3. Bounding box UI scales relative to zoom level

**Expected Behavior**
The bounding box UI (in particular the drag handles) should have a fixed size. Zooming in, not so bad, but zooming out can result in tiny handles.

Jacco:
Call setSelectedItems from zoom functions (Fixes #219)

### DIFF
--- a/src/containers/paint-editor.jsx
+++ b/src/containers/paint-editor.jsx
@@ -36,6 +36,9 @@ class PaintEditor extends React.Component {
             'handleSendToFront',
             'handleGroup',
             'handleUngroup',
+            'handleZoomIn',
+            'handleZoomOut',
+            'handleZoomReset',
             'canRedo',
             'canUndo',
             'onMouseDown',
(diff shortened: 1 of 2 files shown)


paulkaplan (dismissed):
Thanks for the PR! This fixes zooming via the buttons, but doesn't handle zooming via mousewheel. Could you include a fix for that in here as well?

It is probably very similar and I don't want to have to open another issue for that other half of the problem.

Thanks!

Jacco:
Hi Paul,

Thanks for the feedback, didn't even know about this feature :-)

I added one line of code to the paper-canvas.jsx to fix it

Regards, Jacco

paulkaplan:
I added a commit to resolve the conflicts with develop, but this LG I'll merge.

paulkaplan (approved):

PR #337, opened 2018-03-26
Reduce target when fill hole with same color #301

### LINKED ISSUE #301: Grouping and ungrouping this shape makes it lose its colors
groupungroupcolor

1. Draw a circle with the brush
2. Fill it in
3. Make another shape
4. Group the shapes together
5. Ungroup them

Jacco:
Fixes #301

When removing a hole, the parent should be reduced before fillColor is set

### DIFF
--- a/src/helper/group.js
+++ b/src/helper/group.js
@@ -49,7 +49,7 @@ const _ungroupLoop = function (group, recursive, setSelectedItems) {
     // iterate over group children recursively
     for (let i = 0; i < group.children.length; i++) {
         let groupChild = group.children[i];
-        if (groupChild.hasChildren()) {
+        if (groupChild instanceof paper.Group && groupChild.hasChildren()) {
             // recursion (groups can contain groups, ie. from SVG import)
             if (recursive) {
                 _ungroupLoop(groupChild, recursive, setSelectedItems);
--- a/src/helper/tools/fill-tool.js
+++ b/src/helper/tools/fill-tool.js
@@ -121,8 +121,11 @@ class FillTool extends paper.Tool {
                     this.addedFillItem.fillColor.type !== 'gradient' &&
                     this.fillItem.parent.fillColor.toCSS() === this.addedFillItem.fillColor.toCSS()) {
                 this.addedFillItem.remove();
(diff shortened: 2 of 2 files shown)


thisandagain:
Hi @jacco. Thanks for filing! In the future, can you please provide more details with your issues and PRs? You can both issue and PR templates here:
https://github.com/LLK/scratch-gui/blob/develop/.github/ISSUE_TEMPLATE.md
https://github.com/LLK/scratch-gui/blob/develop/.github/PULL_REQUEST_TEMPLATE.md

Jacco:
I didn’t file the issue. But I’m guessing you want some more documentation
on the PR? I did refer to the issue.

I’ll try to use the templates more :-)

Regards Jacco

Op ma 26 mrt. 2018 om 18:04 schreef Andrew Sliwinski <
notifications@github.com>

fsih:
Ah, it's my bad, I have not set up templates on Scratch paint

fsih (changes requested):
It looks like the issue is that `ungroupItems` in `scratch-paint/src/helper/group.js` is running reduce on compound paths with 1 child. Since the child of a compound path doesn't have fillColor information itself, it loses its color.

This PR fixes the fill tool to prevent it from creating compound paths with only 1 child. I think it's a good thing to fix, since there's no need to have compound paths with only 1 child. But the root of the issue is that ungroup is supposed to be run only on groups, not compound paths.

Do you mind changing the line
`if (groupChild.hasChildren()) {`
to
`if (groupChild instanceof paper.Group && groupChild.hasChildren()) {`
in `scratch-paint/src/helper/group.js`? That way if compound paths with 1 child are created any other way, this bug won't come up.

Thanks!

Jacco:
I added the change like you requested.

I saw isGroup() being used a few lines above would that not be better?

`
if (isGroup(groupChild) && groupChild.hasChildren()) {
 `

fsih:
They are pretty similar. instanceof will match if groupChild is a Group or extends Group, and isGroup will only match if it's a Group

fsih (approved):
Looks good!

PR #339, opened 2018-03-26
Fix editorconfig indentation rules

bates64:
Rules before:

* 4-wide tabs everywhere, except
* .js and .html files, using 4 spaces **(note this project also uses .ejs & .jsx files which are not covered by this!)**

Changed in this PR to:

* 4 spaces everywhere (ie. including .jsx and .ejs files), except
* .yml files, using 4-wide tabs
* .json files (eg. package.json), using 2 spaces (may want to change package.json to use 4sp?)

Also fixed index.ejs to use 4 spaces instead of two.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,8 +4,13 @@ root = true
 end_of_line = lf
 insert_final_newline = true
 charset = utf-8
-indent_size = 4
 trim_trailing_whitespace = true
 
-[*.{js,html}]
 indent_style = space
+indent_size = 4
+
+[*.yml]
+ident_style = tab
+
+[*.json]
+indent_size = 2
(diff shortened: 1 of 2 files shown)


thisandagain:
Thanks @heyitsmeuralex !

fsih (changes requested):
The .editorconfig file matches our other repos (e.g. scratch-gui, scratch-vm) so maybe it would make sense to leave them consistent, or change all the repos at the same time?

I think this makes sense, except I don't think we should use tabs for indentation anywhere. Can you switch .yml to 2 space indentation? Thanks!

bates64:
@fsih done as you said :)

fsih (approved):
Cool, thanks!

PR #345, opened 2018-03-29
Text follow ups: don't strip newlines and use Scratch default size/font

paulkaplan:
1. Update the default size/font to correspond with Scratch2.
2. Remove the "fix" for parsing svg files with newlines/tabs, it was fixed in chrome 61 and was breaking multiline text rendering.

See https://github.com/LLK/scratch-svg-renderer/pull/11 for more info on svg text fixes.


cwillisf (approved):
L`GT`M

Issue #332, opened 2018-03-22
Holding shift while rotating text does not do the right thing


fsih:
This will be relevant once the textArea branch is merged.

Shift-rotate a text element:
![shifttext](https://user-images.githubusercontent.com/2855464/37779135-07219f86-2dc2-11e8-9d27-1b9ce41b086f.gif)

Jacco:
Hi there,

I did some research on this one. I haven't found why this only goes wrong with TEXT and not with other SHAPES. But I found a fix for the issue in rotate-tool.js:

The fix also fixes #343 (since that issue is essentially a duplicate?)

Let me know if you want a PR for this one (it did not mention the 'HELP WANTED' label).

```
    onMouseDrag (event) {
        let rotAngle = (event.point.subtract(this.rotGroupPivot)).angle;
        
        for (let i = 0; i < this.rotItems.length; i++) {
            const item = this.rotItems[i];
            
            if (!item.data.origRot) {
                item.data.origRot = item.rotation;
            }
            
            if (event.modifiers.shift) {
                rotAngle = Math.round(rotAngle / 45) * 45;
            }
            item.rotate(rotAngle - this.prevRot[i], this.rotGroupPivot);
            this.prevRot[i] = rotAngle;
        }
    }
```

This code is a little simpler than the old code. Holding SHIFT now only changes the rotAngle and is handled identically with normal rotate.

Another point I found: The prevRot does not need to be an array, in how the routine currently works. If you want this simplified let me know. I can easily add it to the PR.

Regards, Jacco

fsih:
@Jacco Nice! It's much simpler to see what's happening this way. I'm not totally sure why applyMatrix = false was originally added. It looks like it came from here: https://github.com/w00dn/papergrapher/commit/a038207cfc75442cc9ad0b311bb6565bc031b57a#diff-cc810f12347213d487c4055f3382028dR121 so it may have been needed for the rotation handles to work, and in Scratch we aren't using rotation handles currently, so it should be fine to remove.

That's a good point about prevRot too! Please do simplify it. Thanks for all your help!

fsih:
Closed via #359

PR #359, opened 2018-04-03
Issue/332 Rotate-tool fix

Jacco:
### Resolves

Fixes issue #332 concerning text rotation holding shift. Somehow for text items the old code did not work while for other items it did.

### Proposed Changes

The code for shift handling is now simplified and does what it is supposed to do.

Also I changed prevRot to not be an array. It was not needed as an array a number was enough.

### Reason for Changes

Holding shift while rotating seemed to change pivot point while rotating. It was a bug.

### Test Coverage

I did not add tests, but tested rotating several different objects in Chrome.

PS. Sorry for the messy merge PR with the merge commit in it. I know it is possible to make it cleaner and in one commit but I need some advice on how to do it. I probably need to make a new clean fork once all PRs are handled.

### DIFF
--- a/src/helper/selection-tools/rotate-tool.js
+++ b/src/helper/selection-tools/rotate-tool.js
@@ -27,7 +27,7 @@ class RotateTool {
                 this.rotItems.push(item);
             }
         }
-        this.prevRot = 90;        
+        this.prevRot = 90;
     }
     onMouseDrag (event) {
         let rotAngle = (event.point.subtract(this.rotGroupPivot)).angle;
--- a/src/helper/selection-tools/rotate-tool.js
+++ b/src/helper/selection-tools/rotate-tool.js
@@ -31,20 +31,21 @@ class RotateTool {
     }
     onMouseDrag (event) {
         let rotAngle = (event.point.subtract(this.rotGroupPivot)).angle;
-        
+        if (event.modifiers.shift) {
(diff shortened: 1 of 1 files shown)


src/helper/selection-tools/rotate-tool.js
```diff
@@ -44,13 +41,9 @@ class RotateTool {
...
                 rotAngle = Math.round(rotAngle / 45) * 45;
-                item.applyMatrix = false;
-                item.pivot = this.rotGroupPivot;
-                item.rotation = rotAngle;
-            } else {
-                item.rotate(rotAngle - this.prevRot[i], this.rotGroupPivot);
             }
-            this.prevRot[i] = rotAngle;
+            item.rotate(rotAngle - this.prevRot, this.rotGroupPivot);
+            this.prevRot = rotAngle;
```
fsih: I'm seeing an issue where you can only rotate one thing
rotbug

I think you want to bring this line out of the loop

fsih (changes requested):

Jacco:
Good catch. Didn't test that. Code seemed logic :-) Working on it

Jacco:
Hm I dont have that issue it seems :S How do you get the cat in the stand alone scratch-paint?

Jacco:
I do have it now: Its the simplification :P

fsih (approved):
Looks good, thanks!

PR #360, opened 2018-04-03
Add ability to fill transparent text

fsih:
### Resolves
Along with https://github.com/LLK/paper.js/pull/9 , this change allows transparent text to be filled with the fill tool. Before, transparent text would not be detected by the fill tool at all.

### Proposed Changes

For point text, don't require hasFill()

### Reason for Changes

The fill tool should be able to fill in things you've made transparent, including text

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "react-dom": "^16"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


fsih:
poke

paulkaplan (dismissed):

PR #368, opened 2018-04-07
Replace png background with vector version (for #155)

### LINKED ISSUE #155: Zoom blurs background grid
Not super critical, but this isn't the way Scratch 2.0 works - the grid stays sharp when zoomed in there.

Scratch 3.0:

Blurred grid

Scratch 2.0:

Sharp grid

Jacco:
### Resolves

#155 Blurry background when zooming

### Proposed Changes

Use paperjs vector objects to describe background. This is done using one white rectangle, a colored path inside a group which is inserted in the guideLayer instead of background.png image.

The old background.png and its reference are removed since they are no longer needed.

I created a new function for it that accepts hor/ver count of squares and color to accommodate future changes to the grid.

### Reason for Changes

Vectors don't blur when zooming because the are not made of pixels.

### Test Coverage

- I tested the background in Chrome
- The background is not erased by eraser tool


fsih (approved):
This looks great, thanks!

PR #369, opened 2018-04-09
Make sure to remove the bounding box when no items are selected

fsih:
No bounding box should be shown when no items are selected.


paulkaplan (approved):
Ah the good old "truthy empty array" problem... LG

PR #370, opened 2018-04-10
Convert to bitmap

fsih:
### Resolves

Enable the convert to/from vector/bitmap button

### Proposed Changes

- Converting to bitmap flattens the vector content to a new raster layer. Converting back to vector trims the transparent pixels from the raster layer and moves the raster back onto the vector layer.

- Undo/Redo states now need to keep track of which format the art was in

- Changing the format changes the set of tools (currently bitmap has 0 tools.)

- For now, bitmaps export wrapped in SVGs. Having the paint editor take/export bitmaps is a larger change involving more repos that will be part of separate PRs.

### Reason for Changes

So there's a place to put the bitmap tools


paulkaplan:
@fsih it looks like this needs to be merged with develop because of the vector background changes in layer.js

paulkaplan (approved):
This looks good! Like we talked about IRL, we may want to change the "skip_convert" formats, refactoring so that the conversion takes place before the events instead of in response to them. But that might be more complicated because of undo, so this looks ok for now.

Also just note after you merge this we won't be pulling paint changes into GUI until we get to the point where we would put this on preview, specifically getting the bitmap tool icons in with "coming soon".

PR #374, opened 2018-04-11
Convert to bitmap shape rendering

fsih:
Get crisp conversions to bitmap via SVG shape-rendering instead of using paper.js's rasterize

Before:
image

After:
image

UPDATE, now with double resolution:
image

Caveat: It doesn't work to anti-anti-alias text.


src/containers/paper-canvas.jsx
```diff
@@ -82,14 +82,36 @@ class PaperCanvas extends React.Component {
...
+        const bounds = paper.project.activeLayer.bounds;
+        const svg = paper.project.exportSVG({
+            bounds: 'content',
+            matrix: new paper.Matrix().translate(-bounds.x, -bounds.y)
+        });
+        showGuideLayers(guideLayers);
+        
+        // Get rid of anti-aliasing
+        // @todo get crisp text?
+        svg.setAttribute('shape-rendering', 'optimizeSpeed');
```
cwillisf: How does it look with `crispEdges` instead of `optimizeSpeed`? It seems like `crispEdges` is a better semantic match for the intent here, but I'm not sure of the actual behavior.

I did a quick test in Chrome by loading [the Scratch Cat SVG](https://cdn.assets.scratch.mit.edu/internalapi/asset/09dc888b0b7df19f70d81588ae73420e.svg/get/) directly then messing with the SVG in the dev console, and it looks the same with `crispEdges` or `optimizeSpeed`. The [MDN article on `shape-rendering`](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/shape-rendering) seems to imply that it could vary across browsers, though.

src/containers/paper-canvas.jsx
```diff
@@ -82,14 +82,36 @@ class PaperCanvas extends React.Component {
...
+            raster.onLoad = function () {
+                const subCanvas = raster.canvas;
+                getRaster().drawImage(
+                    subCanvas,
+                    new paper.Point(Math.floor(bounds.topLeft.x), Math.floor(bounds.topLeft.y)));
+                paper.project.activeLayer.removeChildren();
+                this.props.onUpdateSvg();
+            }.bind(this);
         }.bind(this);
+        img.src = `data:image/svg+xml;charset=utf-8,${svgString}`;
```
cwillisf: Wow. That's a little bit hilarious, all written out!

It does seem like you could simplify it a bit by replacing the `.bind(this)` stuff with arrow functions; is there a reason not to do so?

cwillisf (dismissed):
Overall it looks good to me. I had a couple of questions that might generate minor changes, but I don't think they should be treated as blocking the merge if you'd like to move on.

src/containers/paper-canvas.jsx
fsih: I think you're right that crisp edges seems more likely to work.

These specs sure use the word "might" a lot...

src/containers/paper-canvas.jsx
fsih: Sounds good. I am just not used to the es6 notation still.

Jacco:
I don't understand why this feature should 'get rid of anti-aliasing'. If you keep the anti aliasing (comment the line) you hardly see difference but it is actually converted to a bitmap (zoom in and you'll see). Getting rid of the anti-aliasing sounds more like a color limiting filter? I would (almost) never need the "after" cat :-)

fsih:
Yeah, it looks pretty choppy, but that's how Scratch 2 works. It's much easier to use with the bitmap fill tool this way. I'm working on increasing the resolution, so after that it will look a bit better when converted to bitmap.

PR #376, opened 2018-04-12
Fill-mode cursor #358

### LINKED ISSUE #358: Replace the cursor with the fill bucket icon when in fill mode
Whenever the fill tool is active and the mouse is over the paint canvas, the usual mouse cursor should disappear, and be replaced with the fill tool icon (located at `scratch-paint/src/components/fill-mode/fill.svg`)

The icon should be lined up with the pointer location like this:
image

/cc @carljbowman

Jacco:
### Resolves

Fixes #358

### Proposed Changes

Added cur regex to webpack loader config
Added map from mode to css classname
Added PaperCanvas has extra css classname bound to props.mode

### Reason for Changes

As requested #358 

### Test Coverage

Tested in Chrome


src/containers/paper-canvas.css
```diff
@@ -5,3 +5,7 @@
     position: absolute;
     background-color: #fff;
 }
+
+.fill-mode {
+	cursor: url(../components/fill-mode/fill-mode.cur) 29 29, auto;
+}
```
fsih, line 11: Add a newline at the end of the file (lint requirement)

src/containers/paper-canvas.jsx
```diff
@@ -264,7 +266,10 @@ class PaperCanvas extends React.Component {
     render () {
```
fsih, line 266: Rather than adding ModeClasses to modes.js, can you add a switch statement on this.props.mode here, to choose the css class string?

modes.js is in lib/ which implies that it should be stand-alone, so it's odd for it to refer to specific CSS in scratch-paint.

fsih (changes requested):
Tested:
Working on Chrome, Windows!
Working on Firefox, Windows!
Not working on Edge, Windows :( Maybe it's this issue: https://github.com/angular/angular-cli/issues/7069
I wouldn't block merging this PR on Edge though; we can file a separate bug to follow up

fsih:
Closing for now as we are on UI change lockdown for the rest of the year, and want to rethink how we will design paint editor cursors. I will come back to this after

PR #382, opened 2018-04-12
Loupe issues fixed #378

### LINKED ISSUE #378: Eyedropper issues: blurriness, picking from checkerboard and guides
### Expected Behavior

No blurriness, colors picked come from drawing only, not guides or checkerboard

### Actual Behavior

ezgif com-video-to-gif

Sometimes the grey and blue are picked (see center square in loupe)

Jacco:
### Resolves

Fixes issues with the loupe #378 

### Proposed Changes

Make a snapshot of svg and paint it in ZOOM_SCALE to a canvas and use that for the loupe display to avoid blurriness. The snapshot is made without guides and background because the user does not want to pick those colours.

### Reason for Changes

I removed the overflow hidden to always see the full circle of the loupe in stead of it moving under the edges of the paper. I used GUIDE_COLOR for the part of the loupe view that does not come from the drawing (near the edge).

Discussion: It should be possible to grab colours with transparent but I thinks transparency is also a missing slider from the color selectors. Also there might be modes: 1) get color from first object you hit (might have transparency), 2) get color from end rendering. It is still possible to pick color that have an anti-aliased value not directly from an object, I think its not would you would expect.

### Test Coverage

I have tested in Chrome


fsih:
@paulkaplan We may want to pull some of this back into the loupe for the pen tool

src/components/loupe/loupe.jsx
```diff
@@ -21,49 +20,50 @@ class LoupeComponent extends React.Component {
...
+        const loupeDiameter = loupeRadius * 2;
 
         const color = this.props.colorInfo.color;
 
         const ctx = this.canvas.getContext('2d');
-        this.canvas.width = ZOOM_SCALE * (LOUPE_RADIUS * 2);
-        this.canvas.height = ZOOM_SCALE * (LOUPE_RADIUS * 2);
+        this.canvas.width = loupeDiameter;
+        this.canvas.height = loupeDiameter;
+        ctx.fillStyle = getGuideColor();
```
fsih: The blue around the edges is unexpected, I think white might be better

src/helper/tools/eye-dropper.js
```diff
@@ -1,11 +1,24 @@
...
 const LOUPE_RADIUS = 20;
+const ZOOM_SCALE = 3;
 
 class EyeDropperTool extends paper.Tool {
     constructor (canvas, width, height, pixelRatio, zoom, offsetX, offsetY) {
         super();
 
+        const guideLayers = hideGuideLayers();
+
+        const colorShot = paper.project.exportSVG({asString: true});
```
fsih: Transforming between the paper format and SVG and back is pretty expensive, so we should avoid it if we can. For instance, it takes almost 5 seconds on the complicated basketball background.

Luckily we're moving data from canvas to canvas, so it shouldn't be too bad. There is a class in paper called Raster (which is basically a canvas) that would work well. Unfortunately, the code is kind of complicated, since paper can only rasterize one layer at a time, and then you have to merge them manually. I have written up some code at https://github.com/fsih/scratch-paint/commit/9ef38cdb7fba4eba2c5c3dc42c16dc2868a90f62 that shows how to do it this with rasters. It took about 100ms to transform the basketball background.

fsih (dismissed):
Looks awesome! I'm excited to get this in

fsih:
I noticed that it looks a bit unusual when the loupe is over the convert button:
image

src/helper/tools/eye-dropper.js
Jacco: Your code looks great :-) Are you going to merge this, or do you want me to do something?

I saw the behavior over the button too... No idea how that works LOL, I thought I got rid of it using the blue but somehow thats semi transparent.

fsih: I have pushed my commit to your branch so you can pull it in.

I think the overlap is caused by the 'Convert to Bitmap' <Button> being on top of the <Loupe> element in scratch-paint/src/components/paint-editor/paint-editor.jsx. It ends up looking see-through because the 'Convert to Bitmap' button has opacity set to .75. It could be fixed by changing the order of the elements so the button is first, and adding some Flexbox CSS to rearrange them, or we could put back 'overflow: hidden' and maybe come back to it later, up to you :)

Jacco:
Ok, thanks but the new version doesn't correctly handle the loupe when
zoomed in. Still needs some work

fsih:
Thanks for catching the zoom issues! Sorry for taking so long to get around to fixing them.
The fixes depend on https://github.com/LLK/paper.js/pull/17

Issue #366, opened 2018-04-06
Make escape key leave text edit mode


fsih:
### Expected Behavior

Pressing escape in scratch 2 moves from text edit mode to select mode
![esc2](https://user-images.githubusercontent.com/2855464/38427372-113fb18a-3987-11e8-8f4d-61f6780fe366.gif)

### Actual Behavior

Escape does nothing in scratch 3

### Help wanted: Where to start
scratch-paint/src/helper/tools/text-tool.js controls the text tool. In this class, this.setTextEditTarget(null) would end text editing mode and make the text selected.

### Operating System and Browser

Windows 10 Chrome

fsih:
Closed via #383

PR #383, opened 2018-04-13
Escape end text edit mode #366

Jacco:
### Resolves

Fixes issue #366

### Proposed Changes

Handle escape key while editing in onKeyDown calling endTextEdit

Discussion: 

1. This probably goes for all tools. Maybe we should look for a generic escape key handling?
2. Is saw that some code is never triggered. 

`if (event.event.target instanceof HTMLInputElement) {`

### Reason for Changes

Issue #366 

### Test Coverage

Tested in chrome

### DIFF
--- a/src/helper/tools/text-tool.js
+++ b/src/helper/tools/text-tool.js
@@ -239,7 +239,9 @@ class TextTool extends paper.Tool {
             // Ignore nudge if a text input field is focused
             return;
         }
-        
+        if (this.mode === TextTool.TEXT_EDIT_MODE && event.key === 'escape') {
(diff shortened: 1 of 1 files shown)


apple502j (changes requested):
Please remove line 244's white spaces after }.
It causes the error.

apple502j (approved):
It seems good.

fsih (approved):
Looks good, thanks!

1. Which other uses of the escape button do you mean?
2. This line is to make sure you are never editing two text fields at once. For instance if you press tab in the text field, then you can get to the sprite name field.

PR #386, opened 2018-04-17
Double resolution of canvas

fsih:
### Resolves

Part of work for supporting Scratch 2's double resolution bitmaps

### Proposed Changes

Double the resolution of the canvas


src/helper/view.js
```diff
@@ -53,7 +56,7 @@ const zoomOnSelection = deltaZoom => {
 };
 
 const resetZoom = () => {
-    paper.project.view.zoom = 1;
+    paper.project.view.zoom = .5;
```
paulkaplan, line 59: I think you also need to change the `math.max` line to be with 0.5 instead of 1 on line 28; zooming out is stuck at half zoom, can't zoom all the way out.

src/containers/paint-editor.jsx
```diff
@@ -214,7 +221,8 @@ class PaintEditor extends React.Component {
             paper.project.view.pixelRatio,
             paper.view.zoom,
             paper.project.view.bounds.x,
-            paper.project.view.bounds.y
+            paper.project.view.bounds.y,
+            isBitmap(this.props.format)
```
paulkaplan: Looks like this is an extra line from a different branch?

src/containers/paper-canvas.jsx
```diff
@@ -104,6 +112,7 @@ class PaperCanvas extends React.Component {
             const raster = new paper.Raster(img);
             raster.onLoad = () => {
                 const subCanvas = raster.canvas;
+                document.body.appendChild(subCanvas);
```
paulkaplan: Doesn't look like this should be here!

paulkaplan (changes requested):
A few cleanup things are needed here.

src/helper/view.js
fsih: Good catch, thanks!

src/containers/paint-editor.jsx
fsih: Done

src/containers/paper-canvas.jsx
fsih: Haha, it's a feature!

paulkaplan (approved):
LG

PR #387, opened 2018-04-17
Bitmap brush

fsih:
### Resolves

Toward adding support for bitmap tools

### Proposed Changes

Adds the bitmap brush tool
bitbrush


src/helper/bitmap.js
```diff
@@ -1,5 +1,85 @@
...
+    let err = dx - dy;
+    
+    callback(x1, y1);
+    while (x1 !== x2 || y1 !== y2) {
+        const e2 = err * 2;
+        if (e2 > -dy) {
+            err -= dy; x1 += sx;
+        }
+        if (e2 < dx) {
+            err += dx; y1 += sy;
```
paulkaplan: probably want only single statement per line!

src/helper/bitmap.js
```diff
@@ -1,5 +1,85 @@
...
+        }
+        callback(x1, y1);
+    }
+};
+
+const fillEllipse = function (centerX, centerY, radiusX, radiusY, context) {
+    // Bresenham ellipse algorithm
+    centerX = ~~centerX;
```
paulkaplan, line 33: https://jsperf.com/tilde-vs-floor/11 I kinda doubt the `~~` is actually faster at this point, but I don't have a super strong opinion ¯\_(ツ)_/¯

src/helper/bitmap.js
```diff
@@ -1,5 +1,85 @@
 import paper from '@scratch/paper';
 
+const line = function (point1, point2, callback) {
```
paulkaplan: Might want to rename this to something like `forEachLinePoint`, since this doesn't actually draw a line.

paulkaplan (dismissed):
A few minor comments that I'll leave up to you. This looks and works great! 

Still having some issues with image getting displaced switching between bitmap and vector, but I know you already have that filed.

src/helper/bitmap.js
fsih: Ok, I took out the comment that said it was faster :p

src/helper/bitmap.js
fsih: Done

src/helper/bitmap.js
fsih: Done

PR #404, opened 2018-04-24
Import and export bitmaps

fsih:
### Resolves
Note that this changes the API to scratch-paint!

Import and export bitmaps. There is a lot of renaming going on here. I have tried to break the commits down so that the functional changes are separate from the renaming, so please look at those.

### Proposed Changes

Previously, the paint editor accepted SVG strings and output SVG strings.
It now accepts bitmaps in the form of HTMLImageElements, and outputs bitmaps in the form of HTMLCanvasElements. These are the formats that happened to work easily with our existing machinery, but I could easily switch to importing and outputing the same thing, if that seems more reasonable.

### DIFF
--- a/src/containers/paper-canvas.jsx
+++ b/src/containers/paper-canvas.jsx
@@ -150,13 +150,7 @@ class PaperCanvas extends React.Component {
     }
     checkFormat (image) {
         if (image instanceof HTMLImageElement) return Formats.BITMAP;
-        if (typeof image === 'string') {
-            const parser = new DOMParser();
-            const svgDom = parser.parseFromString(image, 'text/xml');
-            if (svgDom && svgDom.firstElementChild && svgDom.firstElementChild.tagName === 'svg') {
-                return Formats.VECTOR;
-            }
-        }
+        if (typeof image === 'string') return Formats.VECTOR;
         log.error(`Image could not be read.`);
         return null;
     }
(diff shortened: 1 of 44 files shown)


src/containers/paper-canvas.jsx
```diff
@@ -121,6 +122,7 @@ class PaperCanvas extends React.Component {
         const img = new Image();
         img.onload = () => {
             const raster = new paper.Raster(img);
+            raster.remove();
```
paulkaplan, line 124: What does this do?

paulkaplan (dismissed):
Generally looks good, I had one small question about a line I don't quite understand. 

Also just want to make sure you know that the issue where converting to bitmap/back to vector repeatedly while zoomed in seems to translate and eventually destroy the image.

src/containers/paper-canvas.jsx
fsih: I was seeing an issue where the raster was appearing for a split second at 0, 0 and I thought this might help. I'm not sure if it fixed it... Without this line, the raster would get removed a bit later during paper.project.activeLayer.removeChildren();

fsih:
I will take a look at the converting back and forth issue next

fsih:
On Edge, trying to parse an SVG dom gives permission denied, and was breaking the paint editor. So instead of trying to parse, I added the assumption that all incoming strings are SVG strings.

paper.js's importSVG does its own validation, and invalid SVG strings will render nothing in the editor, so it should be safe.

paulkaplan (dismissed):

fsih:
Ok, this is going in now! Hold off on merging to GUI

PR #405, opened 2018-04-25
Add line tool

fsih:
### Resolves

Adds line tool to bitmap paint editor

### DIFF
--- a/src/helper/bit-tools/line-tool.js
+++ b/src/helper/bit-tools/line-tool.js
@@ -2,11 +2,12 @@ import paper from '@scratch/paper';
 import {getRaster} from '../layer';
 import {forEachLinePoint, fillEllipse} from '../bitmap';
 import {getGuideLayer} from '../layer';
+import {ART_BOARD_WIDTH, ART_BOARD_HEIGHT} from '../view';
 
 /**
- * Tool for drawing with the bitmap brush.
+ * Tool for drawing lines with the bitmap brush.
  */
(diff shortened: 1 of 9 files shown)


src/helper/bit-tools/line-tool.js
```diff
@@ -0,0 +1,142 @@
...
+        // For performance, make sure this is an integer
+        this.size = Math.max(1, ~~size);
+    }
+    // Draw a brush mark at the given point
+    draw (x, y) {
+        const roundedUpRadius = Math.ceil(this.size / 2);
+        this.drawTarget.drawImage(this.tmpCanvas, new paper.Point(~~x - roundedUpRadius, ~~y - roundedUpRadius));
+    }
+    updateCursorIfNeeded () {
+        if (!this.size) {
```
paulkaplan, line 45: This looks like a lot of it is copied from the bit brush tool, could you put this into a helper to avoid the copy/pasting?

paulkaplan (changes requested):
One request for extracting/reusing the cursor preview code between the bit brush and the bit line, since they seem the same?

src/helper/bit-tools/line-tool.js
fsih: I've moved the canvas drawing part into a helper. Is that what you were looking for?

paulkaplan (approved):

paulkaplan:
Great!

Issue #258, opened 2018-01-17
All the sprites are ungrouped by default


apple502j:
### Expected Behavior
Sprites should be grouped

### Actual Behavior
Not so. They're ungrouped by default.
![See it](https://i.gyazo.com/0e5d282b6bd16b2906e5b5994fb4e5dc.gif)
### Steps to Reproduce
1. Add a sprite.
2. Open Paint Editor.
3. Drag something.
### Operating System and Browser
Windows7 SP1 64bit HomePremium
Firefox 57.0.4

fsih:
This was a small change that we made to make the sprites easier to edit. If you would like to use the sprite as a group, you can select the whole sprite and click the 'group' button at the top

apple502j:
I think new scratchers drag the sprite by mistake and cannot fix...

Kenny2github:
@apple502j that's why the "undo" button is there 😉

fsih:
We have decided based on lots of feedback like this to go back to grouping the sprite by default

paulkaplan:
Fixed by https://github.com/LLK/scratch-paint/pull/407

PR #407, opened 2018-04-26
Remove ungroup on import

fsih:
### Resolves
https://github.com/LLK/scratch-paint/issues/258

### Proposed Changes
Don't ungroup sprites when importing


paulkaplan (approved):
LG

PR #412, opened 2018-04-30
Update import and export types

### LINKED ISSUE #384: Convert between bitmap/vector can get cut off
### Expected Behavior

Full sprite should get converted

### Actual Behavior

The view box can get positioned strangely if you convert while zoomed in
screen shot 2018-04-16 at 13 42 21

### Steps to Reproduce

1. Zoom in so the cat is cut off in the paint editor
2. Convert back and forth from vector and bitmap

MacOSX Chrome

fsih:
More changes to scratch-paint as a result of changes to https://github.com/LLK/scratch-vm/pull/1082

- Paint editor takes an image format string as a prop
- Paint editor takes data URIs as bitmaps and outputs ImageData for bitmaps
- Get rid of uses of paper.view.center and changing in and out of zoom levels and replaced with constants, which should fix https://github.com/LLK/scratch-paint/issues/384

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -1,7 +1,7 @@
 [main]
 host = https://www.transifex.com
 
-[experimental-scratch.scratch-paint]
+[scratch-editor.paint-editor]
 file_filter = translations/<lang>.json
 source_file = translations/en.json
 source_lang = en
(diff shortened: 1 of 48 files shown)


fsih:
I'm seeing an issue where if you convert to bitmap, convert back to vector, switch costumes, and switch back, the costume is offset in the renderer (but not in the paint editor). I'm really stuck on this issue :( so I'm posting this even though that's not fixed.

The issue with the costume must be getting exported to the renderer on the convert to vector step, because that's the only time the renderer receives a new SVG. But at that step the renderer still looks right. The costume doesn't get offset until you switch costumes. Where is something changing between those two steps? :/

paulkaplan (approved):
LG!

PR #423, opened 2018-05-03
Hide top row buttons in bitmap mode

fsih:
### Resolves

Hide grouping and send to back/forward buttons when in bitmap mode, to match mock:
https://projects.invisionapp.com/d/main?gclid=Cj0KCQjwkpfWBRDZARIsAAfeXapFYaINVTamyEFOptx6vYV9rsYa3ofFirSdp5wXKSjC1kQWVZFFjVEaArqSEALw_wcB#/console/13730145/286209509/preview
These buttons don't do anything in bitmap mode.

The first commit breaks out the tools into their own component since the paint editor component is getting bloated. The second commit is the UI change.


paulkaplan (approved):
LG

PR #424, opened 2018-05-04
Listen to undo state in fixedTools

fsih:
Top toolbar should listen to undo state so that undo button enabled state updates correctly


paulkaplan (approved):
LG

PR #427, opened 2018-05-07
Use font inlining method from scratch svg render

fsih:
### Resolves
Depends on: https://github.com/LLK/scratch-2.0-coping-strategies/pull/16
Fixes: Converting to bitmap does not preserve font

It turns out that this issue is fixed already in scratch-svg-renderer, so share that code to paint!
For testing, loaded project #220724734


cwillisf (dismissed):
Looks good, pending merge conflict resolution

Issue #330, opened 2018-03-21
Shape center is offset incorrectly on undo


ericrosenbaum:
I saw this with this project:

https://llk.github.io/scratch-gui/develop/#206831002

- do anything that will enable the undo button
- press undo
- blue rectangle is offset so its center is in the corner:

![paint-undo-bug](https://user-images.githubusercontent.com/567844/37738319-05fb27d4-2d2d-11e8-8c4e-69561a82ccde.gif)

fsih:
@ericrosenbaum I noticed this shape also cannot be erased in the paint editor, and its fill shows up as black. How was it created? Perhaps we should generate a more paint-editor friendly backdrop for the tutorial.

ericrosenbaum:
I created the blue rectangle in the Scratch 2.0 paint editor- pretty sure it was in the vector editor.

That project was a prototype for testing the idea of loading a starter project for the tutorial (which we are not currently planning to do).

fsih:
For the other issues with the blue rectangle, I've filed https://github.com/LLK/scratch-paint/issues/432

PR #430, opened 2018-05-08
Queue initial undo snapshot

fsih:
### Resolves

Fixes https://github.com/LLK/scratch-paint/issues/330

### Proposed Changes
Queue undo snapshot on initial SVG load. This allows transforms to finish applying before taking a snapshot

### Test Coverage
Tested on Windows Chrome and Firefox

### DIFF
--- a/src/containers/paper-canvas.jsx
+++ b/src/containers/paper-canvas.jsx
@@ -250,7 +250,10 @@ class PaperCanvas extends React.Component {
                         .subtract(itemWidth, itemHeight));
                 }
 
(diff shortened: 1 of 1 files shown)


paulkaplan:
LG if it fixes it, but do you know the root cause of this? Is it that one of those transforms is async? Or is there some other part of our code that is applying a transform? Seems a bit spooky...

paulkaplan (approved):

fsih:
I looked into it briefly and didn't find the source of the asyncness. :/ I agree it's not great and we should look into it more. There's a clue in Eric's bug report that this jumping background is one made in Scratch 2.0, and fill and erase don't work properly with it, so it's probably an issue specific to whatever that vector object type is.

Issue #371, opened 2018-04-10
Fonts for Paint Editor


carljbowman:
### Overview
We'd like to add the ability to set a font like in 2.0. As apart of this work we'd like to swap out some fonts that will work across more platforms and support more languages.

### Font List
| 2.0 Font                   | New 3.0 Font        | New font Specimen            |
| ------------------- |------------------ | --------------------------|
| Helvetica Bold         | Noto Bold              | https://www.google.com/get/noto/                               |
| Scratch                    | _Being Designed_ | **TBD**                              |
| Donegal                   | Source Serif          | https://typekit.com/fonts/source-serif/details/source-serif-pro-regular |
| Gloria                       | Handlee                 | https://www.fontsquirrel.com/fonts/handlee |
| Marker                     | Knewave                | http://www.1001fonts.com/knewave-font.html#character-map |
| Mystery                   | Griffy                      | http://www.1001fonts.com/griffy-font.html#character-map |
| _Add Pixel font_      | Press Start P2       | https://fonts.google.com/specimen/Press+Start+2P |
| _Add Chinese_        | @thisandagain      | **TBD**                                        |
| _Add Japanese font_ | @thisandagain   | **TBD**                                        |
| _Add Korean font_  | @thisandagain      | **TBD**                                        |

### Font Menu UI
![typinating comencing - edit text mode](https://user-images.githubusercontent.com/3409578/38581643-70c1a1e8-3cdb-11e8-94a3-1bbff817495d.png)

apple502j:
Speaking of Japanese, I suggest Sazanami Gothic
https://osdn.net/projects/efont/
Including Japanese characters (and alphabets)

Edited 2018/04/15UTC
Lite version:
apple502j/SazanamiGothicLite

Licensed under 3 clause BSD(It uses,too!)

thisandagain:
Japanese: `font-family: "ヒラギノ角ゴ Pro W3", "Hiragino Kaku Gothic Pro", Osaka, "メイリオ", Meiryo, "ＭＳ Ｐゴシック", "MS PGothic";`
Chinese: `font-family: "Microsoft YaHei", "微软雅黑", STXihei, "华文细黑";`
Korean: `font-family: 'Malgun Gothic';`

thisandagain:
Ideally we could use Noto (https://www.google.com/get/noto/) but the SC, TC, JP, and KR packages combined are roughly 20MB for a single weight. Because of this I think will need to fall back to system fonts.

[...]

PR #443, opened 2018-05-15
Font tool

fsih:
### Resolves
Resolves https://github.com/LLK/scratch-paint/issues/371

text
### Proposed Changes
Add a font selector dropdown to the text tool in the paint editor

### Reason for Changes
We like fonts

### Test Coverage
Tested on Windows Chrome, Edge. Typing in custom fonts is pretty slow on Edge.


src/components/mode-tools/mode-tools.jsx
```diff
@@ -33,185 +37,446 @@ import straightPointIcon from './icons/straight-point.svg';
...
+                            imgSrc={flipVerticalIcon}
+                            title={this.props.intl.formatMessage(messages.flipVertical)}
+                            onClick={this.props.onFlipVertical}
+                        />
+                    </InputGroup>
+                </div>
+            );
+        case Modes.TEXT:
+            return (
+                <InputGroup>
```
paulkaplan: It looks like the other mode tools are wrapped in a `<div className={classNames(this.props.className, styles.modeTools)}>`, but the text input is not. This seems to be causing the slight "jump" when you choose the font tool:

jumping-font-tool

Wrapping this InputGroup like the other tools fixes that.

src/components/mode-tools/mode-tools.jsx
```diff
@@ -33,185 +37,446 @@ import straightPointIcon from './icons/straight-point.svg';
...
+                                        {this.props.intl.formatMessage(messages.pixel)}
+                                    </span>
+                                </Button>
+                                <Button
+                                    className={classNames(styles.modMenuItem)}
+                                    onClick={this.handleClick}
+                                    onMouseOver={this.handleChangeFontChinese}
+                                >
+                                    <span className={styles.chinese}>
+                                        中文
```
paulkaplan: Is there something else we can show that indicates that this font supports Chinese (ditto for Korean and Japanese)? It was definitely my first question opening the font menu, as I can't identify characters belonging to those different languages. I think this will cause some consternation among kids (although it also piques curiosity!) Maybe just writing `中文 (Chinese)` would help? If nobody else thinks it is a big deal then nevermind. 

/cc @carljbowman

src/components/mode-tools/mode-tools.jsx
```diff
@@ -33,185 +37,446 @@ import straightPointIcon from './icons/straight-point.svg';
...
+        description: 'Name of the pixelated font',
+        id: 'paint.modeTools.pixel'
+    }
+});
+class ModeToolsComponent extends React.Component {
+    constructor (props) {
+        super(props);
+        bindAll(this, [
+            'getFontStyle',
+            'getTranslatedFontName',
```
paulkaplan: Since all these methods are used only for the font dropdown, would you mind resetting this to a functional component and create a separate container/component for doing the event handling for the font dropdown? It would help keep this file a bit more easy to read, and keep the separation between containers/components.

paulkaplan (changes requested):
A few comments that need to be addressed, but generally looking and working great!

There is one bug I haven't been able to consistently reproduce, but here is a gif of it. It seems possible to have a text selection box left behind and reappear at a different time/on a different sprite. We've seen similar bugs before with selection items, do you where to look to track this one down? 

This gif shows the selection box of a different text element from a different costume reappear when making new text.

selection box reappears

fsih:
I have filed https://github.com/LLK/scratch-paint/issues/447 for that bug for now

src/components/mode-tools/mode-tools.jsx
fsih: Good catch, thanks!

src/components/mode-tools/mode-tools.jsx
fsih: Our plan was to not translate these 3 font names. Using 2 languages might make translation tricky. We could use the name of the language and translate that, but @thisandagain and @carljbowman  liked this better I think.

I would say that this way it better shows what the font will look like when you use it, but for the Asian fonts they may look different on different machines.

fsih: That reminds me, the font names try to be in the font style, so if the font names are translated to unsupported characters, they might look weird. For instance you can see the effect of Marker's limited character set here:
image

Ideas for how to handle this? It mostly looks broken for languages where the translation includes the occasional character with an accent.
@chrisgarrity

src/components/mode-tools/mode-tools.jsx
fsih: Done

paulkaplan (dismissed):
LG! Thanks for making those changes. Keep an eye out for that text-tool selection box bug at the smoketest, might be able to find a repro.

src/components/mode-tools/mode-tools.jsx
fsih: We decided to resolve this by not translating any font names

PR #450, opened 2018-05-17
Dont translate fonts and increase default font size

fsih:
We decided not to translate any of the font names, since there could be issues with the font missing foreign letters and not showing up right.

Also, they were a bit too small to see.
image


kchadha (approved):
LG (the `leading` prop is scary.....)

Issue #452, opened 2018-05-17
Paint editor automatically groups things on export


fsih:
### Expected Behavior
Ungrouped things should stay ungrouped

### Actual Behavior
Paint editor export gets grouped
![autogroup](https://user-images.githubusercontent.com/2855464/40200856-592753a2-59eb-11e8-8d36-1851d905f463.gif)

### Steps to Reproduce
1. Make a new costume
2. Draw 2 things
3. Switch costumes
4. Switch back

### Operating System and Browser
Windows 10 Chrome

PR #453, opened 2018-05-17
Ungroup layer groups generated by paper

fsih:
### Resolves
Fixes https://github.com/LLK/scratch-paint/issues/452

### Proposed Changes
Ungroup groups created by Paper. The painting layer is becoming a group.


kchadha (dismissed):
These changes look good.

Has there also been a change to the circle tool recently? On develop it has a default black stroke (and purple fill) but locally it doesn't have any stroke for me..

fsih:
@kchadha you might have ended up in that situation if you clicked on something in select mode? Select mode changes the fill color/stroke color/stroke width to whatever you select, so if you select a line made with the brush tool, then switch to the circle tool, the stroke will be gone.

PR #474, opened 2018-05-23
Fix firefox wrap and opaque text selection

### LINKED ISSUE #463: Highlighting Text completely hides the text in the selection
From @carljbowman

This is unique to Mac Firefox

### LINKED ISSUE #468: Multiline text not working on Firefox
Multi-line selection of text is really wonky. Create a text box, type and add new lines, and then try to select all the text. @carljbowman

Both Windows and Mac Firefox are affected

textareafirefox

fsih:
### Resolves
Fixes https://github.com/LLK/scratch-paint/issues/468
Fixes https://github.com/LLK/scratch-paint/issues/463

### Proposed Changes
white-space:  nowrap was used to prevent wrapping
white-space: nowrap is implemented very differently in Firefox. In most browsers, it causes the text to not wrap except at line breaks. On Firefox, it causes the text to ignore line breaks.

Since whenever the text is changed, the textbox size is updated to exactly fit the text, it should be safe to make the textbox size just slightly wider than the text to always avoid text wrapping.

Additionally, I've put in a somewhat hacky fix for 'In Firefox on Mac the text selection completely obscures the text', by making the text area slightly transparent. It looks like this
screen shot 2018-05-23 at 17 59 44

### Reason for Changes
Fix some text bugs on firefox


paulkaplan (approved):
LG!

PR #490, opened 2018-06-05
Move convert to bitmap into bitmap helper

fsih:
Code clean-up: Move convert to/from bitmap code into the bitmap helper file.

### DIFF
--- a/src/helper/bitmap.js
+++ b/src/helper/bitmap.js
@@ -1,4 +1,6 @@
 import paper from '@scratch/paper';
+import {clearRaster, getRaster, hideGuideLayers, showGuideLayers} from '../helper/layer';
+import {inlineSvgFonts} from 'scratch-svg-renderer';
 
 const forEachLinePoint = function (point1, point2, callback) {
     // Bresenham line algorithm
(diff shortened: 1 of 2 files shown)


src/helper/bitmap.js
```diff
@@ -141,14 +143,72 @@ const getHitBounds = function (raster) {
     return new paper.Rectangle(left, top, right - left, bottom - top);
 };
 
-const trim = function (raster) {
+const _trim = function (raster) {
```
paulkaplan: Above we use the `_` at the end to indicate functions that don't get exported?

src/helper/bitmap.js
```diff
@@ -141,14 +143,72 @@ const getHitBounds = function (raster) {
...
+            onUpdateImage();
+        };
+    };
+    // Hash tags will break image loading without being encoded first
+    img.src = `data:image/svg+xml;utf8,${encodeURIComponent(svgString)}`;
+};
+
+const convertToVector = function (clearSelectedItems, onUpdateImage) {
+    clearSelectedItems();
+    const raster = _trim(getRaster());
```
paulkaplan: Probably worth just inlining the `_trim` function here since it isn't used elsewhere

paulkaplan (changes requested):
Code LG, but I think there are some convention violation/general cleanup with that trim function (see comments)

paulkaplan (dismissed):
LG

I should mention since these functions are being edited that the call to `getSubRaster` in `convertToVector` is throwing on Firefox when you have a blank (0x0) canvas.

Issue #492, opened 2018-06-06
Rectangle tool behaves incorrectly when pressing shift


fsih:
### Expected Behavior

Holding shift should make the rectangle tool draw a square with one corner at the mouse down point

### Actual Behavior

Moving the mouse up causes the rectangle to shift instead of resize
![rectangleshift](https://user-images.githubusercontent.com/2855464/41070159-4dc16cb2-69bf-11e8-9d48-ee96074fbad7.gif)

### Steps to Reproduce

1. Go to the paint editor
2. Select the rectangle tool
3. Hold shift
4. Draw a rectangle
5. Move the mouse up while drawing the rectangle

### Operating System and Browser
Mac Chrome

PR #493, opened 2018-06-06
Fix drawing rectangle while holding shift

fsih:
### Resolves
Fixes https://github.com/LLK/scratch-paint/issues/492

### Proposed Changes

Keep the corner in the same place when drawing a rectangle using shift
rectangleshift2


paulkaplan (approved):
LG

PR #494, opened 2018-06-06
Bitmap rectangle tool

fsih:
### Proposed Changes

Adds bitmap rectangle tool
bitmaprect

Switching between filled and outlined rectangles will be a separate PR.


src/components/mode-tools/mode-tools.jsx
```diff
@@ -89,6 +89,8 @@ const ModeToolsComponent = props => {
     case Modes.BIT_BRUSH:
         /* falls through */
     case Modes.BIT_LINE:
+        /* falls through */
+    case Modes.BIT_RECT:
```
paulkaplan: Seems a bit strange to have this here. It results in an unlabeled, non-functional input in the mode tools bar. Should this PR just leave out the changes in this file, leaving the mode tools area blank?

paulkaplan:
I noticed an issue with the select mode on bitmap rectangles. You can create an "unselectable" part of the costume if you do the following:

1. Be in bitmap mode
2. Click rectangle tool
3. Create a rectangle
4. _With that rectangle selected_, convert to vector
5. The rectangle is not selectable in the vector editor. If there were other bitmap drawings, those are selectable as a single object, but the new rectangle is not part of that raster object.

bitmap-unreachable-objects

It isn't immediately clear why, it seems like `BitmapRectMode#deactivateTool` is being called correctly when you click "Convert to Vector". Maybe it is not being called before some other important thing?

---

Looking into this more, it does look like `convertToVector` in paper-canvas container is being called before the tool has a chance to deactivate, causing the selected rect not to be included in the conversion. Maybe a call to `commitRect` could be added to the onSelectionChanged handler of the bit rect-tool? That would make it so the call to `clearSelection` in `convertToVector` commits the bitmap before it gets converted

src/helper/bit-tools/rect-tool.js
```diff
@@ -0,0 +1,149 @@
...
+    commitRect () {
+        if (!this.rect || !this.rect.parent) return;
+
+        const tmpCanvas = document.createElement('canvas');
+        tmpCanvas.width = getRaster().width;
+        tmpCanvas.height = getRaster().height;
+        const context = tmpCanvas.getContext('2d');
+        context.fillStyle = this.color;
+        drawRect(this.rect, context);
+        getRaster().drawImage(tmpCanvas, new paper.Point());
```
paulkaplan, line 140: Just a question, but why is the temporary canvas needed? Could it be drawn directly on the raster?

src/helper/bitmap.js
```diff
@@ -145,7 +145,133 @@ const trim = function (raster) {
...
+    forEachLinePoint(startPoint, heightPoint, (x, y) => {
+        context.fillRect(x, y, 1, 1);
+    });
+    forEachLinePoint(endPoint, widthPoint, (x, y) => {
+        context.fillRect(x, y, 1, 1);
+    });
+    forEachLinePoint(endPoint, heightPoint, (x, y) => {
+        context.fillRect(x, y, 1, 1);
+    });
+    floodFill(~~center.x, ~~center.y, context);
```
paulkaplan, line 334: Could this helper be simplified by using the `rect.matrix` to transform the context and use `fillRect` always, instead of only in the special case of no rotation? Looks like paper has a utility for that kind of thing, `matrix.applyToContext`. I did some quick testing and this seems to work, including rotation/scaling:
```
const drawRect = function (rect, context) {
    context.save();
    rect.matrix.applyToContext(context);
    context.fillRect(
        ~~(-rect.size.width / 2), // Since rect matrix transforms to center
        ~~(-rect.size.width / 2),
        ~~rect.size.width,
        ~~rect.size.height);
    context.restore(); // Pop back out of the rect matrix transform
    return;
}
```

I'm sure the flood fill helpers will still be needed for the fill bucket, but they might not be needed here! I wonder if something similar could help with the trouble you mentioned scaling/rotating ellipses?

paulkaplan (changes requested):
A few changes requested:
- The comment in `components/mode-tools/mode-tools.jsx`, important because it looks broken.
- Consider the simplification suggested in `helper/bitmap.js`

And that bug with the selected bitmap rect not being correctly saved when converting to vector.

src/components/mode-tools/mode-tools.jsx
fsih: I'll leave it out until drawing rectangle outlines is added.

src/helper/bit-tools/rect-tool.js
fsih: The raster may already have stuff drawn on it. For instance if you drew a rectangle over the cat, then it might just flood fill part of the cat instead of filling in the rectangle.

src/helper/bitmap.js
fsih: The problem with all of the canvas functions is that they create aliasing:
image
Otherwise paper.js's built in rasterize() function would make everything very easy!

The bitmap tools in Scratch 2.0 don't create aliasing, and that makes them much easier to work with with the fill tool and for making pixel art.

fsih:
"I noticed an issue with the select mode on bitmap rectangles. You can create an "unselectable" part of the costume if you do the following:"

@paulkaplan I think I finally got this working! It was surprisingly involved.
- Convert to vector was being called before deactivateTool, which meant the rectangle sort of missed the train to become a vector and was left on the raster layer
- Once I switched the order of those two things, I still had an issue because the format is set to vector before deactivateTool is called. I tried to call updateImage() in componentWillReceiveProps in rect-tool, when this.props.format was still BITMAP, but apparently by then this.props.format in paint-editor (where handleUpdateImage is) is already VECTOR. So I had to add a boolean in paint-editor that basically reverses what format it uses in handleUpdateImage between just before it switches tools and just before it calls convertToVector.

paulkaplan (approved):

PR #499, opened 2018-06-07
Reungroup

fsih:
Ungroup sprites by default again in the paint editor.

We have found that it's very difficult for new users to pose sprites, because they first have to understand the concept of grouping. Therefore we're going back to ungrouping sprites by default.

### DIFF
--- a/src/containers/paper-canvas.jsx
+++ b/src/containers/paper-canvas.jsx
@@ -196,7 +196,13 @@ class PaperCanvas extends React.Component {
                     item.translate(new paper.Point(ART_BOARD_WIDTH / 2, ART_BOARD_HEIGHT / 2)
                         .subtract(itemWidth, itemHeight));
                 }
-                if (isGroup(item) && item.data && item.data.isPaintingLayer) {
+                if (isGroup(item)) {
+                    // Fixes an issue where we may export empty groups
+                    for (const child of item.children) {
+                        if (isGroup(child) && child.children.length === 0) {
+                            child.remove();
+                        }
+                    }
                     ungroupItems([item]);
                 }
 


src/containers/paper-canvas.jsx
```diff
@@ -196,7 +196,13 @@ class PaperCanvas extends React.Component {
                     item.translate(new paper.Point(ART_BOARD_WIDTH / 2, ART_BOARD_HEIGHT / 2)
                         .subtract(itemWidth, itemHeight));
                 }
-                if (isGroup(item) && item.data && item.data.isPaintingLayer) {
+                if (isGroup(item)) {
```
kchadha, line 199: So what was the `item.data && item.data.isPaintingLayer` check from before, and why is it gone now?

kchadha (commented):
@fsih, I have a question in the comments.

src/containers/paper-canvas.jsx
fsih: Good question!

The issue before (if (isGroup(item) && item.data && item.data.isPaintingLayer)) was
1. Open a costume
2. Add an object to the costume
3. Switch costume, switch back
Then the way the paint editor exports the SVG, it takes the "painting layer" that the paint editor created and converts it into a group. Now when it re-imports the SVG, it has an extra group around everything.
4. (if (isGroup(item) && item.data && item.data.isPaintingLayer)) ungroups that group

The way that ungrouping by default works is
1. Open a costume from import/the library. If the top level is all one group, it gets ungrouped.
2. Add an object to the costume
3. Switch costume, switch back
4. The paint editor exported the SVG with a group around it, but the new ungrouping code automatically ungroups it, so we don't need a special case for it.

Ok, but then say we change a step
1. Open a costume from import/the library. If the top level is all one group, it gets ungrouped.
2. **Group everything**
3. Switch costume, switch back
Theoretical 4. Now you would think that the grouped costume got nested into another group which is the layer, and when you go to ungroup it, everything would still be in a group!
Actual 4. No, because of the item.reduce() earlier in the function, the nested groups got flattened into one group

More people reasoning through it to make sure it makes sense is probably a good idea.

kchadha: So does that last problem in the

> Ok, but then say we change a step

section still remain? This PR doesn't change anything about the item.reduce() function, etc.

fsih: No, the last section is to explain that an issue you might expect to happen actually isn't an issue.
Should we VC about it? Maybe I can explain it better

paulkaplan (approved):
LG! We should file an issue for the fact that you can't manually group things and have them stay grouped, but I'm cool with that for now.w

fsih:
Filed https://github.com/LLK/scratch-paint/issues/513 for user-created top-level groups

PR #504, opened 2018-06-08
Fill tool

fsih:
### Resolves
Built on top of
https://github.com/LLK/scratch-paint/pull/494
Here is the diff: https://github.com/fsih/scratch-paint/compare/rectTool...fsih:fillTool?expand=1

Adds the bit fill tool
I also added click and drag to fill and replace color when holding shift, for funsies. If we like it I can add those to the vector fill tool.
bitfill

### Proposed Changes
Add the bitmap fill tool


fsih:
(I was waiting to assign a reviewer after https://github.com/LLK/scratch-paint/pull/494 went in)

src/helper/bit-tools/fill-tool.js
```diff
@@ -0,0 +1,59 @@
...
+        if (event.event.shiftKey) {
+            this.changed = floodFillAll(event.point.x, event.point.y, this.color, context) || this.changed;
+        } else {
+            this.changed = floodFill(event.point.x, event.point.y, this.color, context) || this.changed;
+        }
+    }
+    handleMouseDrag (event) {
+        const context = getRaster().getContext('2d');
+        if (event.event.shiftKey) {
+            this.changed = floodFillAll(event.point.x, event.point.y, this.color, context) || this.changed;
```
paulkaplan, line 42: love this mode, wish it wasn't hidden!

src/helper/bitmap.js
```diff
@@ -275,27 +275,79 @@ const floodFillInternal_ = function (x, y, imageData, newColor, oldColor, stack)
...
+ * @return {Array<int>} Color, a length 4 array
+ */
+const fillStyleToColor_ = function (fillStyleString) {
+    const tmpCanvas = document.createElement('canvas');
+    tmpCanvas.width = 1;
+    tmpCanvas.height = 1;
+    const context = tmpCanvas.getContext('2d');
+    context.fillStyle = fillStyleString;
+    context.fillRect(0, 0, 1, 1);
+    return context.getImageData(0, 0, 1, 1).data;
```
paulkaplan, line 292: there are not enough emoji to contain how clever/bewildering this is. I mean, it works... 🔥

src/helper/bit-tools/fill-tool.js
```diff
@@ -0,0 +1,59 @@
...
+        this.onMouseDown = this.handleMouseDown;
+        this.onMouseDrag = this.handleMouseDrag;
+        this.onMouseUp = this.handleMouseUp;
+
+        this.color = null;
+        this.changed = false;
+        this.active = false;
+    }
+    setColor (color) {
+        this.color = color ? color : 'rgba(0,0,0,0)';
```
rschamp: Seems like we should be importing the default color string from somewhere (even if it's in a constant in this file) rather than hard-coding it.

fsih: It's not quite a default color? In paper.js having a color be null means that it's rendered as transparent. It doesn't work that way with canvas operations though, null becomes black, so if you want transparent you have to give the rgba string. I'll make a constant for it though.

fsih: I did this and merged to head

paulkaplan:
Looks good!

paulkaplan (approved):

rschamp (approved):
🚰 🚰 🚰

PR #507, opened 2018-06-11
Bitmap eraser tool

fsih:
Adds bitmap eraser
biteraser

### DIFF
--- a/src/helper/bitmap.js
+++ b/src/helper/bitmap.js
@@ -86,9 +86,10 @@ const fillEllipse = function (centerX, centerY, radiusX, radiusY, context) {
 /**
  * @param {!number} size The diameter of the brush
  * @param {!string} color The css color of the brush
+ * @param {?boolean} isEraser True if we want the brush mark for the eraser
  * @return {HTMLCanvasElement} a canvas with the brush mark printed on it
  */
-const getBrushMark = function (size, color) {
+const getBrushMark = function (size, color, isEraser) {
     size = ~~size;
     const canvas = document.createElement('canvas');
     const roundedUpRadius = Math.ceil(size / 2);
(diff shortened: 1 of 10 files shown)


src/helper/bitmap.js
```diff
@@ -86,17 +86,19 @@ const fillEllipse = function (centerX, centerY, radiusX, radiusY, context) {
 /**
  * @param {!number} size The diameter of the brush
  * @param {!string} color The css color of the brush
+ * @param {?boolean} isEraser True if we want the brush mark for the eraser
  * @return {HTMLCanvasElement} a canvas with the brush mark printed on it
  */
-const getBrushMark = function (size, color) {
+const getBrushMark = function (size, color, isEraser) {
```
paulkaplan, line 92: seems like the consumer of this could just pass in `white` for the color if they wanted that? Could you just change brush-tool to use `getBrushMark(this.size, this.isEraser ? 'white' : this.color)`?

paulkaplan: Oh nevermind I see the comment about adding the outline as well

fsih: I could, but I want to draw an outline around the white circle for erasers, just as soon as I get the ellipse code in...

src/reducers/bit-eraser-size.js
```diff
@@ -0,0 +1,31 @@
+import log from '../log/log';
+
+const CHANGE_BIT_ERASER_SIZE = 'scratch-paint/eraser-mode/CHANGE_BIT_ERASER_SIZE';
+const initialState = 20;
```
paulkaplan: I noticed that this is different from the eraser size in vector mode. Is that on purpose?

fsih: No, that's a mistake

paulkaplan (dismissed):
LG! Works great. Feel free to merge after the eraser size change

PR #515, opened 2018-06-14
Bit text tool

fsih:
Adds the text tool in bitmap mode. It uses the exact same tool as vector, I've just added a step where when you exit both text and selection mode, and the format is bitmap, the vector is transferred to the bitmap layer.

There was an important bug to fix, which was that undoing into a text-edit state left the text unselected. This was fine in vector, but in bitmap it leaves around a vector that it's not obvious isn't part of the bitmap yet. So I've added logic to make text selected when you undo into a state where the text is still vector.

### DIFF
--- a/src/components/bit-text-mode/bit-text-mode.jsx
+++ b/src/components/bit-text-mode/bit-text-mode.jsx
@@ -1,27 +1,26 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 
-import {ComingSoonTooltip} from '../coming-soon/coming-soon.jsx';
 import ToolSelectComponent from '../tool-select-base/tool-select-base.jsx';
 
 import textIcon from './text.svg';
 
-const BitTextComponent = () => (
-    <ComingSoonTooltip
-        place="right"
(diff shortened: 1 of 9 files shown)


paulkaplan (changes requested):
There seems to be a bug with the undo stack in text. Staying just in vector mode, if you add text, then add a new block of text ,and another one, they are not separate undos.

fsih:
Undo in vector mode should be fixed.

In order to make debugging easier, I passed in a prop to TextMode as we discussed so that only one is active at a time.

Last time we talked about rearranging the code in on mouse down the next time I touched this code, since it was hard to understand. I tried to simplify the code by adding more early returns and adding beginSelect and endSelect to match beginTextEdit and endTextEdit, so hopefully it's easier to see what's going on, and that also helped with debugging how saving undo state was failing.

paulkaplan (dismissed):
LG! Thanks for refactoring some of that logic. Again, let's make sure to get plenty of eyes on these changes, since they involve changing both vector and bitmap text tools

PR #516, opened 2018-06-15
Bitmap Oval tool

fsih:
### Resolves
Adds bitmap oval tool


paulkaplan:
Something seems to be wrong with the commit code, it positions the oval at the top left every time when it commits.

fsih:
Hi Paul, it looks like you fixed the issue you mentioned. Is there more for me to do here before you review?

paulkaplan (approved):
As far as I can tell this LG, but obviously there is a ton of math here. Let's just make sure this gets plenty of usage at the smoketest.

Issue #524, opened 2018-06-25
Bug will filling filled rotated bitmap rectangles


fsih:
### Expected Behavior
Rectangle should look the same after clicking away

### Actual Behavior
The rectangle may not be filled, or the entire screen may be filled
![bitrectfill](https://user-images.githubusercontent.com/2855464/41860886-c1fb4276-786d-11e8-936e-1f1559032e35.gif)
![bitrectfill2](https://user-images.githubusercontent.com/2855464/41860889-c433b0fa-786d-11e8-9d6b-ef64f5a0588f.gif)

### Steps to Reproduce
1. Go to paint editor
2. Switch to rectangle tool
3. Make a rotated rectangle
4. Move it so its center is off the editor

### Operating System and Browser
all

PR #525, opened 2018-06-25
Redo fill bitmap rectangle

fsih:
### Resolves
Fixes https://github.com/LLK/scratch-paint/issues/524

### Proposed Changes
Rewrite the rectangle drawing algorithm to not use flood fill. This version calculates the lines through all the edges and fills in the space between.

### Reason for Changes
Flood fill did not work when the rectangle was partially off the edge of the screen. It also didn't work for very narrow rectangles where the interior could get broken into multiple areas.

### DIFF
--- a/src/helper/bit-tools/rect-tool.js
+++ b/src/helper/bit-tools/rect-tool.js
@@ -1,6 +1,6 @@
 import paper from '@scratch/paper';
 import Modes from '../../lib/modes';
-import {drawRect} from '../bitmap';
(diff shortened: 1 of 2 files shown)


paulkaplan (approved):
Seems good, but there might be some minor fixes to the rounding in certain edge cases

fsih:
I've filed https://github.com/LLK/scratch-paint/issues/533 but I think I'm going to leave that bug for now, since I want to get to the other tools

PR #532, opened 2018-06-26
Fix scrolling speed on Firefox

### LINKED ISSUE #529: Scrolling with the mouse wheel is very slow in Firefox
### Expected Behavior
Mouse wheel should scroll at about the same speed in all browsers

### Actual Behavior
Firefox scrolls much more slowly

### Steps to Reproduce
Scroll in the paint editor in Firefox

Windows Firefox

towerofnix:
### Resolves

Fixes #529.

### Proposed Changes

Code changes basically copied from LLK/scratch-blocks#1427:

> Adds a "line-scroll multiplier" value, which is multiplied when the delta mode is line (0x1) instead of pixel (0x0). This makes scrolling on Firefox bearable (as well as on other browsers which might report such a delta mode).

### Reason for Changes

To fix scrolling on Firefox!

### Test Coverage

Manually tested. Uses the same multiplier as the PR in scratch-blocks, which makes scrolling one "unit" of my scroll wheel go the same distance in Firefox and Chromium (where this isn't an issue).

/cc @fsih (made the issue, helped with [a previous scrolling-related issue](https://github.com/LLK/scratch-paint/pull/263), @paulkaplan (assigned to [the scratch-blocks PR](https://github.com/LLK/scratch-blocks/pull/1427/))

### DIFF
--- a/src/containers/paper-canvas.jsx
+++ b/src/containers/paper-canvas.jsx
@@ -176,7 +176,7 @@ class PaperCanvas extends React.Component {
 
                 ensureClockwise(item);
                 scaleWithStrokes(item, 2, new paper.Point()); // Import at 2x
-                
+
                 if (typeof rotationCenterX !== 'undefined' && typeof rotationCenterY !== 'undefined') {
                     let rotationPoint = new paper.Point(rotationCenterX, rotationCenterY);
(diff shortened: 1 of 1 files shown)


fsih (approved):
Awesome, thanks!

PR #537, opened 2018-06-28
Bitmap select tool

fsih:
Adds bitmap select tool

select

### DIFF
--- a/src/components/bit-select-mode/bit-select-mode.jsx
+++ b/src/components/bit-select-mode/bit-select-mode.jsx
@@ -1,27 +1,25 @@
 import React from 'react';
-
-import {ComingSoonTooltip} from '../coming-soon/coming-soon.jsx';
+import PropTypes from 'prop-types';
 import ToolSelectComponent from '../tool-select-base/tool-select-base.jsx';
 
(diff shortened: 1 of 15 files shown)


paulkaplan:
It looks like copy/paste causes a crashing error when used between bitmap and vector. Repro:
- Starting in vector, select the whole costume and hit "copy"
- Convert the costume to bitmap, hit "paste"
- All seems ok, there is a duplicate that is selected, you can move it around.
- Clicking outside doesn't seem to commit the vector to bitmap, leaves the vector object which is still selectable in the bitmap select tool
- Sometimes if you have that pasted object selected, choosing a different bitmap tools crashes the editor.

paulkaplan (changes requested):
The code looks good overall. The only issue is that you can paste vector content into the bitmap editor (possibly vice-versa?) Once that is fixed then the rest looks good.

fsih:
That paste issue should be fixed

paulkaplan:
Looks good! I'm going to go ahead and merge this so that it is available for testing today.

paulkaplan (approved):

fsih:
Yay, thank you Paul!!

PR #545, opened 2018-07-05
Delete button

### LINKED ISSUE #277: Not able to delete selection on touch device
You can select things, but cannot delete them on a touch device like an iPad.

/cc @carljbowman

paulkaplan:
### Resolves

Resolves https://github.com/LLK/scratch-paint/issues/277

### Proposed Changes

Add a delete button that has the same functionality as the backspace keyboard handler, i.e. can delete selections as well as points. 

image


fsih:
screen shot 2018-07-06 at 11 06 25

The delete button breaks the appearance of the adaptive UI for me. This is especially a bummer since both the adaptive UI and delete button are needed for tablet screens. The eventual solution is to have the words fill and outline move to the top of the indicators, and having buttons move to the more menu. Maybe as a short-term solution we can hide the flip buttons when the screen size is small?
I think @chrisgarrity may be actively working on the 'stuff doesn't fit' issue

fsih:
screen shot 2018-07-06 at 11 13 16

Sometimes in the reshape tool, the delete button doesn't activate when it should
1. When you have something selected in the select tool, then switch to the reshape tool. Delete should delete the selected shape. 
2. When you select a point in the reshape tool, then click "curved". Delete should delete the selected point.

paulkaplan:
Fixed the delete button disabled state in 20228df, good catch.

paulkaplan:
@fsih I made the text labels wrap above the color inputs, which freed up enough space for the delete button on small screen. There are still those awkward intermediate pixels where things get too large and stack vertically, but that isn't related to this PR, and I'd like to go through those more carefully.

src/components/mode-tools/mode-tools.jsx
```diff
@@ -151,18 +157,29 @@ const ModeToolsComponent = props => {
...
+                    <LabeledIconButton
+                        disabled={!props.hasSelectedUnpointedPoints}
+                        imgSrc={straightPointIcon}
+                        title={props.intl.formatMessage(messages.pointed)}
+                        onClick={props.onPointPoints}
+                    />
+                </InputGroup>
+                <InputGroup className={classNames(styles.modLabeledIconHeight)}>
+                    <LabeledIconButton
+                        disabled={!(props.hasSelectedUncurvedPoints ||
```
fsih: `!props.selectedItems.length` should suffice, an item is selected if any of its points are selected

fsih (dismissed):
LG besides the comment

PR #549, opened 2018-07-10
Correct the view bounds when the paper canvas is initialized.

fsih:
View bounds is initialized to the identity matrix, but paper actually starts out at 50% zoom (due to having to handle double resolution bitmaps). This would cause updateViewBounds to not trigger its listeners if you pressed the zoom in button (which actually does bring the zoom to %100)


paulkaplan (approved):
LGTM

PR #550, opened 2018-07-10
Draw oval and rectangle outlines in bitmap

fsih:
### Resolves
Draw oval and rectangle outlines in bitmap

For reviewer: there is a lot going on here. I have reorganized the logical chunks of works into commits, so reading the last 5 commits one by one should help.

### Proposed Changes
Ovals and rectangles have a filled/outlined toggle


src/components/mode-tools/mode-tools.jsx
```diff
@@ -234,6 +250,46 @@ const ModeToolsComponent = props => {
...
+                <InputGroup>
+                    <LabeledIconButton
+                        highlighted={!props.fillBitmapShapes}
+                        imgSrc={outlineIcon}
+                        title={props.intl.formatMessage(messages.outlined)}
+                        onClick={props.onOutlineShapes}
+                    />
+                </InputGroup>
+                <InputGroup>
+                    <Label text={props.intl.formatMessage(messages.thickness)}>
```
paulkaplan: Looks like thickness should only be shown if you are on the not filled state

src/helper/bit-tools/oval-tool.js
```diff
@@ -57,17 +57,49 @@ class OvalTool extends paper.Tool {
      */
     onSelectionChanged (selectedItems) {
         this.boundingBoxTool.onSelectionChanged(selectedItems);
-        if ((!this.oval || !this.oval.parent) &&
+        if ((!this.oval || !this.oval.isInserted()) &&
                 selectedItems && selectedItems.length === 1 && selectedItems[0].shape === 'ellipse') {
             // Infer that an undo occurred and get back the active oval
             this.oval = selectedItems[0];
-        } else if (this.oval && this.oval.parent && !this.oval.selected) {
```
paulkaplan, line 64: What is the change to the if logic involving `isInserted` do?

fsih: I'm not sure how, but I hit an issue where the oval had a parent, but it was a group that had no parent, so it wasn't attached to the view. I discovered isInserted in the API and it's exactly what I wanted, that the item is in the view. I might go through later and change more occurrences of parent !== null to isInserted.

paulkaplan: ah cool

paulkaplan (dismissed):
LGTM

src/components/mode-tools/mode-tools.jsx
fsih: Fixed

Issue #449, opened 2018-05-17
Edits that only delete nodes using the reshape tool are lost


ericrosenbaum:
### Expected Behavior

Any edit should update the stage and thumbnails, and persist.

### Actual Behavior

Edits with the reshape tool that only involve deleting one or more nodes do not trigger an update and are lost when you switch away from the paint tab:

![paint-delete-node-bug](https://user-images.githubusercontent.com/567844/40188413-dc0ea686-59c7-11e8-8ea8-a2107cefac97.gif)

### Steps to Reproduce

See gif

### Operating System and Browser

Mac OS 10.13.2 Chrome 66

PR #551, opened 2018-07-11
Update image when deleting points

fsih:
### Resolves
Fixes https://github.com/LLK/scratch-paint/issues/449

### Proposed Changes
Call onUpdateImage when deleting points

### Reason for Changes
Point deletes were not saved


paulkaplan (approved):
LGTM

PR #553, opened 2018-07-12
Make the bitmap/vector conversion button look enabled

### LINKED ISSUE #2589: Convert to bitmap/vector in paint editor looks disabled
### Expected Behavior

Expected a brighter blue button indicating it's clickable

### Actual Behavior

image

### Steps to Reproduce

preview.scratch.mit.edu -> Costumes

_e.g. Mac OS 10.11.6 Safari 10.0_

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-gui/issues/2589

### Proposed Changes

Don't change the opacity of this button, it is always enabled. Remove the active state, which seems like it was made in error, probably mistaken for `enabled`. But now the button is always enabled, so just don't modify the opacity.

### Reason for Changes

It looked disabled.

### Test Coverage

image

/cc @thisandagain

### DIFF
--- a/src/components/paint-editor/paint-editor.css
+++ b/src/components/paint-editor/paint-editor.css
@@ -140,12 +140,6 @@ $border-radius: 0.25rem;
     font-weight: bold;
     color: white;
     justify-content: center;
-    opacity: .75;
-}
-
-.bitmap-button:active {
-    background-color: $motion-primary;
-    opacity: 1;
 }
 
 .bitmap-button-icon {


thisandagain (approved):

fsih:
I'm worried that because it's so bright and large, it looks like you're supposed to click it (like an ok button on an alert). Convert to vector/bitmap isn't super important to call action to in the paint editor UI. Could we make it a white button instead?

@carljbowman do you have thoughts?

fsih:
The button is 75% opacity by default so that it can brighten responsively when you click on it

Issue #363, opened 2018-04-04
Gradient fills for bitmap editor


fsih:
![image](https://user-images.githubusercontent.com/2855464/38321800-9ae2c5bc-3806-11e8-8901-c858ddcab5b9.png)

PR #558, opened 2018-07-16
Vector gradient

fsih:
gradients
(Thanks MysticUnicornSparkle for project 144885018!)

### Resolves
Fixes https://github.com/LLK/scratch-paint/issues/541
Fixes https://github.com/LLK/scratch-paint/issues/362

### Proposed Changes
Adds gradient UI

### Testing
Up at https://llk.github.io/scratch-gui/gradients/

PR #559, opened 2018-07-16
Bitmap gradient

fsih:
### Resolves
Fixes https://github.com/LLK/scratch-paint/issues/363
Depends on https://github.com/LLK/scratch-paint/pull/558
Diff here: https://github.com/fsih/scratch-paint/compare/vectorGradient...fsih:bitmapGradient?expand=1

### Proposed Changes
Gradient fill bucket in bitmap mode

### Test Coverage
Up at https://llk.github.io/scratch-gui/gradients/


fsih:
Fixes since smoke test (we might want to test these fixes):
- Changing the gradient type always replaced the second color with a generated color. This should now only happen when changing from solid to a gradient. (Fixed in both vector and bitmap editors)
- When zoomed in, filling with bitmap gradients should work as expected
- The bitmap fill tool now makes an undo state every time any color change occurs. It used to batch color changes in click and drags. Now those will be separate. This fixes an issue where the renderer wouldn't pick up color changes to gradients, because those are asynchronous.
- On HDPI screens, the radial gradient center was offset from the click location

paulkaplan (approved):
LGTM.

PR #560, opened 2018-07-16
Adjust format for different locales

### LINKED ISSUE #543: Handle paint editor buttons in other languages
The text under the icons for  group/ungroup, move forward/backward etc, doesn't fit in the space in many languages. Approaches:
* truncate text with '...' and have full text in a tooltip (issue - is the truncate function locale aware)
* remove text and just have a tooltip
* keep text as is, and move buttons that don't fit into a drop down menu (this might be best, but figuring out which languages need to have more buttons in the menu on a case by case basis may not be maintainable in the long run)

/cc @fsih

chrisgarrity:
### Resolves

Resolves #543 

### Proposed Changes

* increased fullsize editor width, English required 1272px, adding a couple more pixels allowed a couple of other languages to keep labels
* hide flip vertical/horizontal labels for everything except English
* hide all labels for languages defined in `hideLabel`

### Reason for Changes

### Test Coverage

Current tests run - further testing will need to happen in gui with all the languages.

### DIFF
--- a/src/components/fixed-tools/fixed-tools.jsx
+++ b/src/components/fixed-tools/fixed-tools.jsx
@@ -19,6 +19,7 @@ import Label from '../forms/label.jsx';
 import LabeledIconButton from '../labeled-icon-button/labeled-icon-button.jsx';
 import {isVector} from '../../lib/format';
 import layout from '../../lib/layout-constants';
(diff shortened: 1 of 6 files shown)


paulkaplan:
@chrisgarrity this looks good, once that center-align commit comes in, merge away

Issue #566, opened 2018-07-20
Bitmap select tool deletes selection when changing costumes


paulkaplan:
1. Make a rectangular selection in bitmap mode
2. Move the selection or rotate it.
3. Change costumes
4. The portion of the previous costume that was selected gets deleted!

Possibly related is that it is confusing at step 2 when the selection is modified, it disappears from the stage/thumbnail.

PR #569, opened 2018-07-25
Save bitmap selection

fsih:
### Resolves
Fixes https://github.com/LLK/scratch-paint/issues/566

### Proposed Changes
Export the bitmap selection when exporting the image to storage

### Reason for Changes
This way the image doesn't appear to disappear from the stage when part is selected.


paulkaplan (dismissed):
LGTM

PR #573, opened 2018-07-31
Localize tool tooltips

### LINKED ISSUE #562: Localize tool tooltips
### Expected Behavior

_Please describe what should happen_
The tooltips (title) for tools like 'circle', 'text', 'rectangle' should be extracted for localization.

### Actual Behavior

Passing the FormatMessage descriptor as a prop is not recognized by the tools as a localization string definition, so it is not extracted. Any strings created this way are not available to translators for translation.
```
<ToolSelectComponent
        imgDescriptor={{
            defaultMessage: 'Circle',
            description: 'Label for the oval-drawing tool',
            id: 'paint.ovalMode.oval'
        }}
        imgSrc={ovalIcon}
        isSelected={props.isSelected}
        onMouseDown={props.onMouseDown}
    />
```
### Steps to Reproduce

_Explain what someone needs to do in order to see what's described in *Actual behavior* above_
Switch to any language other than English (e.g., Italian, French, Spanish) the tooltips are not translated.

If you log on to transifex and search for 'circle' in the Scratch Editor, it is not found.

apple502j:
### Resolves
#562 

### Proposed Changes
Localize tool tooltips so that we can translate on Transifex.

### Reason for Changes
Currently, we cannot translate them on Transifex.

### DIFF
--- a/src/components/bit-brush-mode/bit-brush-mode.jsx
+++ b/src/components/bit-brush-mode/bit-brush-mode.jsx
@@ -11,7 +11,7 @@ const messages = defineMessages({
         description: 'Label for the brush tool',
         id: 'paint.brushMode.brush'
     }
-}); 
+});
 
 const BitBrushModeComponent = props => (
     <ToolSelectComponent
--- a/src/components/bit-brush-mode/bit-brush-mode.jsx
+++ b/src/components/bit-brush-mode/bit-brush-mode.jsx
@@ -2,7 +2,6 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import ToolSelectComponent from '../tool-select-base/tool-select-base.jsx';
 import {messages} from '../../lib/messages.js';
-import {defineMessages} from 'react-intl';
 
 import brushIcon from './brush.svg';
 
--- a/src/components/bit-brush-mode/bit-brush-mode.jsx
+++ b/src/components/bit-brush-mode/bit-brush-mode.jsx
@@ -1,17 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import ToolSelectComponent from '../tool-select-base/tool-select-base.jsx';
+import {messages} from '../../lib/messages.js';
 import {defineMessages} from 'react-intl';
 
 import brushIcon from './brush.svg';
-
-const messages = defineMessages({
-    brush: {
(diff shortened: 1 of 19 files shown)


src/components/bit-brush-mode/bit-brush-mode.jsx
```diff
@@ -1,16 +1,21 @@
...
 import brushIcon from './brush.svg';
 
+const messages = defineMessages({
+    brush: {
+        defaultMessage: 'Brush',
+        description: 'Label for the brush tool',
+        id: 'paint.brushMode.brush'
+    }
+});
+
```
chrisgarrity: @apple502j this is looking really great. There's one slight issue - the same `id` should not be defined in multiple places. Currently the definitions are all the same, but it can lead to bugs later when only one gets changed. To fix this the messages that are shared between components should be pulled out into a shared file. Probably it should be something like `.../lib/messages.js`. It just needs to import `defineMessages` and export the messages object: 
```
import {defineMessages} from 'react-intl';

const messages = defineMessages({
    brush: {
        defaultMessage: 'Brush',
        description: 'Label for the brush tool',
        id: 'paint.brushMode.brush'
    },
    eraser: {
        defaultMessage: 'Eraser',
        description: 'Label for the eraser tool',
        id: 'paint.eraserMode.eraser'
    },
    etc....
});

export default messages;
```

Then in the component, just import messages from the shared file instead of defining the message.

To find duplicates you can run `npm run i18n:src` it will stop as soon as it finds one. Just remember to remove the temporary translations folder if it stops with an error.

chrisgarrity (changes requested):
The paint editor runs into problems with defineMessages because it has identical strings in multiple components. While it's not a problem for the javascript, it's something that can cause problems when synchronizing with Transifex.

src/components/bit-brush-mode/bit-brush-mode.jsx
apple502j: Thanks! I'll try it.

apple502j:
I don't know why lint warned - It's default export.

chrisgarrity (approved):
Thanks @apple502j this looks great. Technically the ones that are just in the vector editor (e.g. reshape) could have been left in the original file because they're not duplicated, but I think having all the tools messages in one place was the correct choice.

Issue #461, opened 2018-05-22
Double clicking text with the Select tool does not allow you to edit the text. It should transition to the Text tool and set the state of the text element to edit.


fsih:
Also when text is selected with the Select tool and you click to the Text tool, the selected text element loses selection.

from @carljbowman

fsih:
Add this to direct select tool as well

PR #583, opened 2018-08-06
Double click in select modes switches to text tool

fsih:
### Resolves
Fixes https://github.com/LLK/scratch-paint/issues/461

### Proposed Changes
Double clicking on a text item in select or reshape mode switches you to the text tool

This implementation has the somewhat weird side effect that if you have a text item and only a text item selected in select mode, and then you switch to the text tool by clicking on the text tool, it still enters text edit mode on the box. Does that seem reasonable?

### DIFF
--- a/src/containers/reshape-mode.jsx
+++ b/src/containers/reshape-mode.jsx
@@ -45,7 +45,8 @@ class ReshapeMode extends React.Component {
             this.props.clearHoveredItem,
             this.props.setSelectedItems,
             this.props.clearSelectedItems,
-            this.props.onUpdateImage
+            this.props.onUpdateImage,
+            this.props.switchToTextTool
         );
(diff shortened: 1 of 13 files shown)


paulkaplan:
Lint is failing

kchadha (approved):
I think this PR looks good.

Re: your comment in the PR description -- if a single text item is selected in select mode and then you manually switch to the text tool, the text goes into edit mode -- that behavior didn't seem weird to me.

One thing we discussed offline, which is worth noting, is that you need to double click on actual rendered text instead of anywhere within the selection bounding box. I think we should maybe file an issue for this and see if it comes up during smoke testing as well.

kchadha:
Actually, I can't seem to repro that last thing I mentioned in my previous comment... O.o

fsih:
Ok thanks! Yeah, let's see if those issues come up in smoke testing.

PR #597, opened 2018-08-09
Reduce hit tolerance

### LINKED ISSUE #144: Difficult to move shape without resizing it
@fsih reported:
> I keep resizing when I'm trying to move (horizontal elongation)

@jwzimmer reported:
> Move a small object. I keep stretching it instead of moving it : (

fsih:
### Resolves
Fixes https://github.com/LLK/scratch-paint/issues/144

### Proposed Changes
- Reduce hit tolerances
- Don't allow hits on helper items that aren't handles. This fixes an issue where after clicking on an item, it gets deselected

### Reason for Changes
Hit tolerances are currently too high. If a rectangle this large is drawn:
image
then the hit area of the handles is actually double the radius of the visible hit targets:
image
making it really difficult to click on the area in the center of the shape in order to move it (you'll probably resize it by mistake.) This resulted when we increased the size of the hit handles, but forgot to decrease the tolerances to match.

If you did manage to hit the center, you would probably deselect the shape instead of being able to move it, which is what the second fix is for.

### Test Coverage
To test:
1. It should now be possible to move small rectangles without reshaping them, as long as you can see a space between the handles in the middle

Things to make sure aren't broken:
1. Handles should still work. At any zoom level, if you are clicking on the handle (the little bubble), then the handle should be selected and draggable.
2. The rotate icon is a handle too. It should not be difficult to use the rotation handle.
3. Check bitmap/vector select tools, as well as reshape tool

### DIFF
--- a/src/helper/bit-tools/oval-tool.js
+++ b/src/helper/bit-tools/oval-tool.js
@@ -11,7 +11,7 @@ import NudgeTool from '../selection-tools/nudge-tool';
  */
 class OvalTool extends paper.Tool {
     static get TOLERANCE () {
-        return 6;
+        return 2;
     }
     /**
      * @param {function} setSelectedItems Callback to set the set of selected items in the Redux state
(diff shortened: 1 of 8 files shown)


kchadha (approved):
LG

PR #616, opened 2018-08-21
Mirror margins, borders and arrow icons for RTL

### LINKED ISSUE #574: Make paint work for RTL languages
### Expected Behavior

Everything should be laid out correctly when using an RTL language such as Hebrew.
- [ ] buttons should have the rounded corners on the 'outside':
screen shot 2018-08-06 at 4 58 44 pm
screen shot 2018-08-06 at 5 00 16 pm
- [ ] labeled inputs should have margin mirrored
- [ ] undo/redo arrows will need to be mirrored
- [ ] button icons need margin reversed:
screen shot 2018-08-06 at 5 02 22 pm
- ~~[ ] editing text with the text tool should put the cursor in the correct place~~ (split out to #615)

Splitting out from https://github.com/LLK/scratch-gui/issues/1579

### Actual Behavior

Layouts specific to LTR look broken on RTL - e.g. margin on the wrong side.

chrisgarrity:
I verified that things like the paint brush and magnifying glass should not be mirrored (even people who read Hebrew are usually right-handed)

### Resolves

Resolves #574 

### Proposed Changes

Changes styles based on whether there is a `[dir="rtl"]` or `[dir="ltr"]` on an ancestor of the current style.

Renamed a couple of styles from 'right' and 'left' to 'start' and 'end' so that they continue to make sense in RTL as well as LTR.

### Testing

- [ ] tool layouts should be RTL in Hebrew
- [ ] undo/redo icons should be mirrored in Hebrew
- [ ] other icons (e.g. paint tools) are still right-handed (not mirrored)
- [ ] button borders should have corner radii in the correct places
- [ ] dashed borders between buttons should mirror
- [ ] margins on elements should be mirrored

### DIFF
--- a/src/components/color-button/color-button.css
+++ b/src/components/color-button/color-button.css
@@ -12,8 +12,16 @@
     flex-shrink: 0;
     height: 100%;
     border: 1px solid rgba(0, 0, 0, 0.25);
+}
+
+[dir="ltr"] .color-button-swatch {
     border-top-left-radius: 4px;
     border-bottom-left-radius: 4px;
+}
+
+[dir="rtl"] .color-button-swatch {
+    border-top-right-radius: 4px;
+    border-bottom-right-radius: 4px;
 }
 
 .color-button-arrow {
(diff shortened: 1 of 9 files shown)


fsih (dismissed):
Just one thing, I think the little arrow next to 'more' should be sticking to the left side, but it's right next to the text:
image

fsih (approved):

BryceLTaylor:
Looks Good!

PR #617, opened 2018-08-23
Make PaintEditor component rtl aware

### LINKED ISSUE #614: Mirror color picker in the paint editor for RTL languages
Splitting out from #574 

Color picker is a modal outside of the gui component, so it doesn't pick up the `dir="rtl"` attribute from gui. Paint is going to need it's own way to determine whether the current locale is RTL

/cc @fsih

chrisgarrity:
Fixes the color-picker and any other popover elements.

### Resolves
Resolves #614 

### Proposed Changes

Adds an RTL prop to the PaintEditor that initializes the `layout` state in redux. Any other components that need to know the layout refer to the state in redux.

I debated whether the state should just be a boolean (true for RTL), or ‘rtl’, ‘ltr’. I went with the latter, but could be convinced that boolean would be better.

I did not add `dir=“rtl”` to the font picker dropdown as all the names are in LTR languages.

Question: Should the sliders reverse direction, and if so, is it worth doing right now when the layout of the color picker may change.

Adding the rtl prop and the `dir=…` to the PaintEditorComponent fixes layout issues in the playground.

### Test Coverage

screen shot 2018-08-23 at 9 15 44 am

screen shot 2018-08-23 at 9 16 49 am

- [ ] Color picker layout reverses (but not the sliders)
- [ ] Color picker icon gradient preview reverses direction (in the swatch next to the dropdown icon)
- [ ] Dropdown menus are mirrored
- [ ] Font menu does NOT mirror

### DIFF
--- a/src/components/color-button/color-button.css
+++ b/src/components/color-button/color-button.css
@@ -12,16 +12,14 @@
     flex-shrink: 0;
     height: 100%;
     border: 1px solid rgba(0, 0, 0, 0.25);
-}
-
-[dir="ltr"] .color-button-swatch {
     border-top-left-radius: 4px;
     border-bottom-left-radius: 4px;
 }
 
+/* mirror the whole swatch so that the colors in the gradient match the
+direction of the color picker */
 [dir="rtl"] .color-button-swatch {
-    border-top-right-radius: 4px;
-    border-bottom-right-radius: 4px;
+    transform: scaleX(-1);
 }
 
 .color-button-arrow {
--- a/src/components/color-button/color-button.css
+++ b/src/components/color-button/color-button.css
@@ -12,14 +12,16 @@
     flex-shrink: 0;
     height: 100%;
     border: 1px solid rgba(0, 0, 0, 0.25);
+}
+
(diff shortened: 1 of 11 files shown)


fsih (commented):
rtl looks good in the playground, but when integrated with GUI locale=he looks like this for me now (paint editor direction is ltr):
image

fsih:
Wait, it isn't doing that now... maybe there's a race?

fsih:
Oh wait, I realized I need to change GUI too

fsih (commented):
Ok new review, now that I've fixed GUI

image
The preview of the gradient doesn't match what's actually applied when an item is selected or filled with the fill bucket. I'm not sure if the fill/select tools should change (that it fills right to left when you're in a right to left language, but if you switch languages, then suddenly the colors you have selected change) or if the two colors should not change sides, and in RTL there will be the weird behavior that the main color becomes the "second" color when switching to a gradient.

fsih (approved):
Looks good! You might want to hold off on merging the paint update to GUI until GUI is updated to provide the rtl prop to the paint editor

Issue #216, opened 2017-12-13
Use zoom-to-fit as the default size for editing a costume


chrisgarrity:
When entering the paint-editor with no zoom level set, it would be nice if the zoom level was set to that the costume fills the editor. When switching between costumes in the editor the zoom-level shouldn't change (to support animation).

From observation in a workshop - kids were happier once they had the costume at a zoom level where it was easy to see the handles. Usually that was not the default zoom.

paulkaplan:
@carljbowman @fsih this needs a bit more specification before it can be implemented.

1. We specifically don't reset the view of the paint editor when switching costumes because it makes it hard for people doing animations to have the view constantly reset. So when specifically do you want the view to zoom-to-fit? On sprite change? This is the only complicating part of this, because right now the costume editor doesn't know anything about sprites, just which costume it is loading.
2. How specifically do we want the artwork to fit within the view? I assume not out to the edges, because that is too close, but we should zoom-to-fit tightly and then pull back a bit. This doesn't have to be decided in advance, I can demo a few different options.

carljbowman:
@paulkaplan - Sorry for the delay. Here are some of my initial thoughts.

> 1. We specifically don't reset the view of the paint editor when switching costumes because it makes it hard for people doing animations to have the view constantly reset. So when specifically do you want the view to zoom-to-fit? On sprite change? This is the only complicating part of this, because right now the costume editor doesn't know anything about sprites, just which costume it is loading.

It does sound like this is a bit more complicated than anticipated. I do think it's extremely important to maintain the view state (zoom + scroll position) when navigating costumes within a sprite.

As of now, I could see two paths forward, both of which still require a sprite to define it's own default view state: a zoom level based on a sprite's costume(s) with the largest width and height (scroll position is not relevant in the default state).  

1a. Switching sprites would cause a reset to the default zoom level and scroll position for all of the sprite's costumes. Zooming and scroll would act like (2.0 & 3.0) and be persistent across the sprite's costumes such that we retain the ease of animating.

1b. Sprites would retain their zoom level and scroll position when navigating sprite to sprite and only appear in the default view state when a sprite/costume is added. 

> 2. How specifically do we want the artwork to fit within the view? I assume not out to the edges, because that is too close, but we should zoom-to-fit tightly and then pull back a bit. This doesn't have to be decided in advance, I can demo a few different options.

[...]

PR #629, opened 2018-08-28
Zoom to fit

fsih:
### Resolves
Fixes https://github.com/LLK/scratch-paint/issues/216
Demo at https://llk.github.io/scratch-gui/zoomToFit/

### Proposed Changes
- Zoom to fit when first opening a sprite in the costume editor
- Zoom level is saved per-sprite
zoomtofit

### Reason for Changes
Editing to-scale sprite is difficult, hopefully this will make it easier


src/reducers/zoom-levels.js
```diff
@@ -0,0 +1,49 @@
...
+        });
+    default:
+        return state;
+    }
+};
+
+// Action creators ==================================
+const saveZoomLevel = function (zoomLevel) {
+    if (!(zoomLevel instanceof paper.Matrix)) {
+        log.warn(`Not a matrix: ${zoomLevel}`);
```
chrisgarrity, line 31: Should this continue to save a zoom level if it's not a matrix?

fsih:
I made the rename and documentation changes we discussed.

chrisgarrity (dismissed):
Looks good!

src/reducers/zoom-levels.js
fsih: Hmm, I'm not sure how to fail gracefully here. I don't think I can not return an action

fsih:
I'm going to hold off on merging because I think the idea is that we'll try this on its own branch during the smoke test, and then decide whether to merge it afterwards.

Issue #605, opened 2018-08-15
When nothing is selected have Copy and Delete act on all artwork


carljbowman:
### Overview
Currently when nothing is selected, clicking Flip Horizontal or Flip Vertical will cause the action to effect all the artwork within a Costume or Backdrop. 

Both Copy and Delete actions feel like good candidates for a similar UX. 
- Delete would act like Scratch 2.0's Clear feature.
- Modifying Copy would mean someone would not have to drag-select everything to Copy / Paste it — as we have seen drag selecting can be challenging for new comers.

### Delete Flows (similar to Scratch 2.0's Clear)
**Delete via UI**
1. Have none of the artwork selected 
2. Delete UI would appear enabled
3. Clicking Delete would delete all artwork within a Costume or Backdrop

**Delete Key**
Deleting everything if you hit the delete key might feel jarring, so let's not optimizing the keyboard input to clear everything.

### Copy Flows
**Copy via UI**
1. Have none of the artwork selected
2. Copy UI would appear enabled
3. Clicking Copy would copy all of the artwork within a Costume or Backdrop

**Copy via Hotkeys**
1. Have none of the artwork selected
2. Using the Copy hotkeys would copy all of the artwork within a Costume or Backdrop

/cc @ntlrsk

carljbowman:
@fsih - Updated the original comment to include the UX that we discussed for hotkeys. WOO! PAINT EDITOR HOTKEYS!

fsih:
Filed https://github.com/LLK/scratch-paint/issues/653

PR #634, opened 2018-08-30
Copy and delete work on all objects when none are selected

fsih:
### Resolves
https://github.com/LLK/scratch-paint/issues/605

### Proposed Changes
When nothing is selected, copy button should copy everything and delete button should delete everything.
Ctrl+c copies everything when nothing is selected, but delete/backspace continue to do nothing if nothing is selected.

### Reason for Changes
Less disabled buttons is more gooder

### Test Coverage
Test copy button, delete button, copy shortcut, delete shortcuts.


src/components/mode-tools/mode-tools.jsx
```diff
@@ -192,7 +191,6 @@ const ModeToolsComponent = props => {
                 </InputGroup>
                 <InputGroup className={classNames(styles.modLabeledIconHeight)}>
                     <LabeledIconButton
-                        disabled={!props.selectedItems.length}
```
ericrosenbaum, line 195: Looking at this together, we decided that the copy button should be disabled if there's no artwork (e.g. after deleting everything)

ericrosenbaum (commented):
Looks good pending one fix to copy button state

src/components/mode-tools/mode-tools.jsx
fsih: I was looking into this and it's more work than I thought, because we aren't really keeping track in the state right now of anything that could indicate that the canvas is empty. I think we should file it as a separate issue as you suggested at first.

fsih: Filed https://github.com/LLK/scratch-paint/issues/653

PR #642, opened 2018-08-31
Disable keyboard shortcuts for now

fsih:
Keyboard shortcuts have too many bugs, so we will disable them for now

Disable keyboard shortcuts that start with meta keys
Ignore delete events when a text field is focused


BryceLTaylor (approved):
Looks good

PR #645, opened 2018-09-01
Make gradient icons not selectable

### LINKED ISSUE #640: Gradient icons should not be selectable
Gradient icons in the color picker become black when you choose one of the gradient types @kathymakes @carljbowman
image36

What's happening here is the icons are getting selected, and the select color is weird because the icons are grayscaled. They just need to have `user-select: none;` css applied to them.

kyleplo:
### Resolves

Resolves #640

### Proposed Changes

Makes gradient icons not selectable.

### Reason for Changes

Prevents accidentally selecting the icons.

### DIFF
--- a/src/components/color-picker/color-picker.css
+++ b/src/components/color-picker/color-picker.css
@@ -98,7 +98,7 @@
     flex-direction: row;
     justify-content: center;
     margin: 8px;
-    user-select: nonr;
+    user-select: none;
 }
 
 [dir="ltr"] .gradient-picker-row > img + img {
--- a/src/components/color-picker/color-picker.css
+++ b/src/components/color-picker/color-picker.css
@@ -98,6 +98,7 @@
     flex-direction: row;
     justify-content: center;
     margin: 8px;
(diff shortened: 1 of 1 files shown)


src/components/color-picker/color-picker.css
```diff
@@ -98,6 +98,7 @@
     flex-direction: row;
     justify-content: center;
     margin: 8px;
+    user-select: nonr;
```
fsih: Thanks!
Looks like there is a typo here in 'none'

It would be good if you could run the scratch-paint playground to make sure the change is working, to catch stuff like this.

fsih (changes requested):

src/components/color-picker/color-picker.css
kyleplo: Fixed. I'm not good at running the playground.

fsih: Ok, let me know if you need help setting it up

fsih (approved):
LGTM!

Issue #615, opened 2018-08-21
Handle RTL languages in the text tool


chrisgarrity:
Splitting out from #574 

- When typing in a RTL language, the textarea should grow to the left (it currently grows to the right)
- When selecting text in an existing RTL textarea the cursor should appear where the mouse was clicked
- If all the text is selected (e.g. by double clicking), the highlight should appear over the text (currently appears offset to the right)
- it should be possible to select a substring of a RTL textarea

@fsih I think we need to pair on this one.

fsih:
@chrisgarrity happy to pair whenever :+1:

PR #651, opened 2018-09-04
Rtl text tool

fsih:
### Resolves
https://github.com/LLK/scratch-paint/issues/615

### Proposed Changes
Correct text behavior in RTL
(The text used to appear in a random place and the highlight and cursor were nowhere near the text)
Test http://localhost:8078/playground/?dir=rtl

### DIFF
--- a/src/containers/text-mode.jsx
+++ b/src/containers/text-mode.jsx
@@ -34,20 +34,25 @@ class TextMode extends React.Component {
         }
     }
     componentWillReceiveProps (nextProps) {
-        if (this.tool && nextProps.colorState !== this.props.colorState) {
-            this.tool.setColorState(nextProps.colorState);
-        }
-        if (this.tool && nextProps.selectedItems !== this.props.selectedItems) {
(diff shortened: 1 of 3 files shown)


rschamp (approved):
This looks good, and thank you for documenting the transforms with comments.

I did notice odd behavior with punctuation going to the wrong side. I think that can be handled separately?

rtlpunctuation

rschamp:
Maybe the deal with the cursor and punctuation is just quirks of using the paint playground in not-completely-RTL mode?

fsih:
I think that's HTML trying to do its best ):

You can start typing English letters in an RTL language, and then you expect them to be added to the right. But ? appears in both English and Hebrew, so if you switch back to Hebrew it thinks you want to start adding letters to the left again. But then you keep typing English letters, so it's like uhh I guess that was a LTR question mark and repositions it. The cursor is wherever the RTL text would appear at all times, because RTL is being considered dominant by the text field.

PR #657, opened 2018-09-06
Unmount deactivate

### LINKED ISSUE #595: Bitmap Shapes disappear if you don't click off of them and then switch costumes
In bitmap, if you make a circle or square and don't click off of it, it doesn't show up on stage and it disappears if you switch costumes and back

fsih:
### Resolves
This fixes https://github.com/LLK/scratch-paint/issues/595 in some cases

### Proposed Changes
On component unmount, deactivate tools.
This way, if you have uncommitted changes, like a bitmap circle or rectangle, switching tabs will commit the change. (This does not fix committing the shape when changing costumes however)

### Reason for Changes
User does not expect the drawn shape to disappear

### DIFF
--- a/src/components/paint-editor/paint-editor.css
+++ b/src/components/paint-editor/paint-editor.css
@@ -142,10 +142,6 @@ $border-radius: 0.25rem;
     margin-left: calc(2 * $grid-unit);
 }
 
-.hidden {
-    display: none;
-}
-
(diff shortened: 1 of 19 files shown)


rschamp (approved):

upupli:
only success on vector mode

Issue #535, opened 2018-06-27
Tools with transparent/no fill in bitmap draw black


hyperobject:
### Expected Behavior
In the bitmap editor, painting with color set to transparent in should behave the same as the erase tool, and drawing shapes with fill transparent should not draw anything (or erase an area that matches the shape drawn).

<img src="https://user-images.githubusercontent.com/3019167/41975980-d65878a6-79e9-11e8-9ec2-ac4e143d0eda.gif" width="300px" />

### Actual Behavior
Using the paint tool or drawing ovals and rectangles draws with black instead of transparent/erasing areas of a drawing.
<img src="https://user-images.githubusercontent.com/3019167/41976064-0f2dc06e-79ea-11e8-8dfe-35cf4ef15507.gif" width="300px" />

### Steps to Reproduce
In the Scratch 3.0 editor, open a bitmap drawing and set fill to transparent. Draw with the paint, oval, or rectangle tools.

### Operating System and Browser

Chrome OS 66, macOS 10.13.5 Firefox Nightly 63

fsih:
Plans for transparent fill
- The brush should become an eraser
- Line, rect, and oval should have transparent color missing from fill menu

msjessicagamble28-dev:
ftcytfy

PR #660, opened 2018-09-07
Handle transparency in bitmap tools

fsih:
### Resolves
Fixes https://github.com/LLK/scratch-paint/issues/535
Fixing this as a prereq for https://github.com/LLK/scratch-paint/pull/662

### Proposed Changes
Take out the option to have transparent fill in line, rect, oval, and text tools in bitmap. Make transparent brush act like eraser.

### Reason for Changes
Transparency in these tools was not supported in scratch 2 and the behavior is weird right now.

### Test Coverage
Test that you can't get into the transparent color in those 4 tools in bitmap mode.
Test that the bitmap brush works and acts like an eraser when set to transparent.


ericrosenbaum (approved):
LGTM!

PR #662, opened 2018-09-11
Export bit shapes

fsih:
### Resolves
Fixes https://github.com/LLK/scratch-paint/issues/595
Depends on https://github.com/LLK/scratch-paint/pull/660 going in first

### Proposed Changes
Like we did for the select tool, oval and rect tool temporary state is exported with bitmaps. This way, they don't disappear if you leave the paint editor before they are committed.

### Reason for Changes
Shapes unexpectedly disappear


ericrosenbaum (dismissed):
👍

Issue #166, opened 2017-10-30
Fencing content to the workspace - (Scrolling phase 1)


carljbowman:
**TL;DR** - Fence artwork to the cursor like Scratch 2.0. More iterations to come later.

## Overview

Currently there are no scrollbars within the vector editor. One can zoom in and navigate with panning gestures, but there is no UI to navigate with.

In discussing ways to fix this issue, we dove down a rabbit hole of dynamic sizing workspaces relative to artwork size. This was to compensate for a related issue of losing artwork. It's easy to drag arwork beyond the boundies of the default zoom "workspace".

### First Phase

We are exciting to explore a more dynamic workspace with artboards in the future, but for now we would like to scope back the work.

- Add Scrollbars when zoomed
- The max that someone can zoom-out should be 1x (100%)
- Artwork should be fenced to the viewport relative the cursor (or touch drag target)

#### Rough Terminology 

These should be consider loose terms for sake of conversation. Properties can shift as needed.

- **Artwork**: Any point, object, pixel that has been created or imported
- **Artboard**: A zoomable fixed space where Artwork can be made. Mainly the Artboard acts as a stage sized reference. Artwork is not limited by this space and can extend beyond.
  - Artwork that extends beyond the Artboard should not be cropped
- **Viewport**:  A scrollable container for the Artboard that masks Artwork/Artboard
- **Fence**: A property of a container that creates a "physical" limitation to where Artwork can go
  - In Phase 1: The Fence will be a property of the Viewport and use the location of the Cursor to define the limitation of where artwork can go. The Scratch 2.0 Paint Editor has an example of this. The Stage also has a concept of a Fence but the limitation is relative to the size of a Sprite.
  - Ultimately the Fence aims to mitigate losing of Artwork by dragging it beyond the visible boundaries of the Artboard, while allowing Artwork to extend the Artboard (e.g Backdrops)

All of these elements interact with each other to determine: size, scale (which relates to the browser size), zoom, and scroll position. Below are a few (but not limited to) examples of how these elements might interact with each other.

*All measurements are rough estimates for the sake of demonstrating relationships.*

![fenced to cursor](https://user-images.githubusercontent.com/3409578/36849924-508772b2-1d33-11e8-9de4-500e077ec5ac.png)

P.S. - This can be fun to simulate with two pieces of paper: one with a rectangular hole cut and one with some artwork. Zooming-in is a bit harder to simulate that zooming out and scrolling.

fsih:
[...]

PR #671, opened 2018-09-11
Fence tools to make it less easy to lose work off the sides

fsih:
### Resolves
Fixes https://github.com/LLK/scratch-paint/issues/166

### Proposed Changes
Fence the move, move point (reshape), nudge, and scale tools to the canvas area.
This also adds super-nudge

### Reason for Changes
Currently, it's easy to move shapes off the edges of the paint editor, which then become inaccessible but are still visible in the thumbnail and on stage, which can be confusing. These changes don't completely prevent moving items off the edges, but make it a more difficult state for beginners to get into.

### Testing
Try moving shapes off the paint editor bounds in the various tools in which you can move things, such as select, reshape, and shape tools, in bitmap and vector. Super-nudge is shift+arrow keys.

### DIFF
--- a/src/helper/selection-tools/move-tool.js
+++ b/src/helper/selection-tools/move-tool.js
@@ -99,12 +99,10 @@ class MoveTool {
     }
     onMouseDrag (event) {
         const point = event.point;
-        if (event.point.x > ART_BOARD_WIDTH || event.point.y > ART_BOARD_HEIGHT ||
-                event.point.x < 0 || event.point.y < 0) {
-            point.x = Math.max(0, Math.min(point.x, ART_BOARD_WIDTH));
-            point.y = Math.max(0, Math.min(point.y, ART_BOARD_HEIGHT));
-        }
+        point.x = Math.max(0, Math.min(point.x, ART_BOARD_WIDTH));
+        point.y = Math.max(0, Math.min(point.y, ART_BOARD_HEIGHT));
         const dragVector = point.subtract(event.downPoint);
+
         for (const item of this.selectedItems) {
             // add the position of the item before the drag started
             // for later use in the snap calculation
--- a/src/helper/selection-tools/move-tool.js
+++ b/src/helper/selection-tools/move-tool.js
@@ -3,6 +3,7 @@ import Modes from '../../lib/modes';
 import {isGroup} from '../group';
 import {isCompoundPathItem, getRootItem} from '../item';
 import {snapDeltaToAngle} from '../math';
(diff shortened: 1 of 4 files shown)


src/helper/selection-tools/move-tool.js
```diff
@@ -97,7 +98,13 @@ class MoveTool {
         this.setSelectedItems();
     }
     onMouseDrag (event) {
-        const dragVector = event.point.subtract(event.downPoint);
+        const point = event.point;
+        if (event.point.x > ART_BOARD_WIDTH || event.point.y > ART_BOARD_HEIGHT ||
+                event.point.x < 0 || event.point.y < 0) {
+            point.x = Math.max(0, Math.min(point.x, ART_BOARD_WIDTH));
```
kchadha: Do you need the big if statement given that you're doing `Math.max` and `Math.min`?

src/helper/selection-tools/nudge-tool.js
```diff
@@ -20,26 +23,40 @@ class NudgeTool {
...
+            translation = new paper.Point(0, Math.min(nudgeAmount, ART_BOARD_HEIGHT - rect.top - 1));
         } else if (event.key === 'left') {
-            translation = new paper.Point(-nudgeAmount, 0);
+            translation = new paper.Point(-Math.min(nudgeAmount, rect.right - 1), 0);
         } else if (event.key === 'right') {
-            translation = new paper.Point(nudgeAmount, 0);
+            translation = new paper.Point(Math.min(nudgeAmount, ART_BOARD_WIDTH - rect.left - 1), 0);
         }
 
+
```
kchadha: stray newline

src/helper/selection-tools/point-tool.js
```diff
@@ -144,7 +145,16 @@ class PointTool {
         this.invertDeselect = false;
         this.deleteOnMouseUp = null;
         
-        const dragVector = event.point.subtract(event.downPoint);
+        const point = event.point;
+        if (event.point.x > ART_BOARD_WIDTH || event.point.y > ART_BOARD_HEIGHT ||
+                event.point.x < 0 || event.point.y < 0) {
+            point.x = Math.max(0, Math.min(point.x, ART_BOARD_WIDTH));
```
kchadha: same question here as above

kchadha: (and same w/the scale tool below as well)

src/helper/selection-tools/move-tool.js
fsih: Oh, I suppose not

kchadha (approved):
LGTM!

Issue #612, opened 2018-08-20
Scrollbars are invisible on Edge


fsih:
### Steps to Reproduce
1. Go to paint editor
2. Zoom in
3. Scrollbars are click and draggable but not visible

Edge is not recognizing 8 letter hex codes as a color. The fix should just be to convert #BEBEBECD to rgba

### Operating System and Browser
Windows Edge

PR #673, opened 2018-09-12
Use RGBA instead of hex so Edge can understand the color

fsih:
### Resolves
Fixes https://github.com/LLK/scratch-paint/issues/612

### Proposed Changes
Make scrollbars visible in Edge

### Reason for Changes
Edge can't understand 8 character hex colors


kchadha (approved):
LGTM

Issue #676, opened 2018-09-12
Filled/outlined and thickness state can get out of sync with the selected item


fsih:
### Expected Behavior
The mode tools row should match the state of the selected item

### Actual Behavior
Filled/outlined and thickness state in bitmap shape editor can get out of sync
![image](https://user-images.githubusercontent.com/2855464/45450070-071cff00-b6a5-11e8-8225-8e9bad597e11.png)

### Steps to Reproduce
1. Switch to bitmap editor
2. Use oval tool to draw an oval
3. Move the oval
4. Switch to outlined
5. Click off the oval
6. Undo

You now have a filled oval but the mode bar still says outlined
If you click off of the oval now, you will get an outlined oval, contrary to what it looks like it will become.

### Operating System and Browser
All

PR #678, opened 2018-09-13
Bitmap select state

fsih:
### Resolves
Fixes https://github.com/LLK/scratch-paint/issues/676

### Proposed Changes
Mode tool state should always match selected item state. When undoing, outline state and thickness update. Outline and thickness changes also update the image on stage.

### Reason for Changes
State not matching can cause issues like the stage not looking like the what's in paint and the wrong thing being drawn when you click off of the shape to commit it.

### Testing
Make a bunch of changes to bitmap shapes, and undo and redo. Make sure that the state in the toolbar matches the state of the shape.


ericrosenbaum (approved):
LGTM!

Issue #663, opened 2018-09-11
Delete in paint editor asks if you want to stay or leave on the page (i.e. delete invokes the back button behavior in the browser). Chrome doesn’t.


fsih:
### Expected Behavior
Delete should not trigger navigate when it does something else

### Actual Behavior
In some browsers, delete tries to navigate

### Steps to Reproduce
Select a shape in paint and delete

### Operating System and Browser
Seen on Firefox Mac

PR #679, opened 2018-09-13
Don't navigate away when paint editor is open and delete/backspace are pressed

fsih:
### Resolves
https://github.com/LLK/scratch-paint/issues/663

### Proposed Changes
Capture backspace and delete key actions when in the paint editor (so Firefox won't try to navigate away from the page)

### Reason for Changes
Delete is a shortcut in Paint and it's frustrating that a pop-up comes up every time, or if you accidentally navigate away from the page.


kchadha (approved):
LGTM

Issue #675, opened 2018-09-12
Bitmap text disappears if you don't click off of it and then switch costumes


fsih:
Same issue as https://github.com/LLK/scratch-paint/issues/595, but for the bitmap text tool.

PR #680, opened 2018-09-13
Export bit text

fsih:
### Resolves
Fixes https://github.com/LLK/scratch-paint/issues/675

### Proposed Changes
Export bitmapified text if there is text being edited when switching costumes

### Reason for Changes
Don't lose bitmap text if you change costumes in the middle of editing text


ericrosenbaum (approved):
LGTM!

Issue #655, opened 2018-09-06
Should dragging to resize selection default to fixed aspect ratio?


ericrosenbaum:
### Expected Behavior

In the 2.0 paint editors, dragging a selection by a corner handle resizes the art but preserves the aspect ratio of the selection (you can dragging the handles on the side of the selection to squash and stretch it).

Should 3.0 match this behavior?

### Actual Behavior

In 3.0, currently dragging a selection by a corner lets your resize and change the aspect ratio, (you can hold shift to preserve the aspect ratio).

fsih:
I think it makes sense to match Scratch 2.0 here, the change was kind of arbitrary. I think @towerofnix suggested somewhere that shift could be used for the reverse, to make resizing stop preserving aspect ratio.

PR #681, opened 2018-09-14
Switch fix aspect ratio to default when resizing by a corner

fsih:
### Resolves
Fixes https://github.com/LLK/scratch-paint/issues/655

### Proposed Changes
Preserve aspect ratio when resizing from a corner
Shift switches to mean don't preserve aspect ratio

### Reason for Changes
Matches 2.0


ericrosenbaum (approved):
LGTM!

PR #682, opened 2018-09-14
Add safety checks to fill and stroke colors

fsih:
### Resolves
Fixes issue found in smoke test:
In Bitmap while drawing a square or circle and using the eyedropper to pick transparency from the backdrop crashes the paint editor.

### Proposed Changes
Perform a safety check before calling toCSS on fill and stroke colors. This reverts the behavior to the behavior on beta, where a transparent object is drawn, and when you click off of it it becomes black.

### Reason for Changes
Fix crash

### Test Coverage
Tested drawing transparent and non-transparent filled/outlined circles and ovals on Windows 10 Chrome, Firefox, and Edge


BryceLTaylor (approved):
Looks Good!

Issue #465, opened 2018-05-22
Adding a very thick outline to text can be rendered differently between the costume editor and the stage.


fsih:
From @thisandagain
![image](https://user-images.githubusercontent.com/2855464/40382385-5158fd6c-5dcc-11e8-9a09-b88a567eca88.png)
The outline in paint looks doubled?

PR #685, opened 2018-09-17
Don't scale text outlines

fsih:
### Resolves
Fixes https://github.com/LLK/scratch-paint/issues/465

### Proposed Changes
Unlike for other shapes, text outline thickness changes with its scale. Since the scale already captures the outline thickness change, there's no need to re-double it in scaleWithStrokes

### Testing
Test loading text with outlines and adjusting text with outlines. The paint editor and stage should match.


kchadha (approved):
LGTM

Issue #520, opened 2018-06-21
Hover outline doesn't necessarily match the next item you will select


fsih:
### Expected Behavior

The outline should indicate what you will interact with if you click

### Actual Behavior
Here the rectangle is hovered but the semicircle is moved
![hoveritem](https://user-images.githubusercontent.com/2855464/41728738-8116e8b2-7545-11e8-828d-219f3f9a0903.gif)

### Steps to Reproduce
1. Be working with lots of shapes on top of each other
2. Click on a big shape that's on top of other things
2. You can hover over shapes behind that shape. You will usually end up moving the shape that's on top instead.

### Operating System and Browser
all

Issue #204, opened 2017-11-20
Can't select front shape with the reshape tool if the back shape is selected


chrisgarrity:
### Expected Behavior

When using the reshape tool, clicking on a shape in front of the selected one should select the front shape.

### Actual Behavior
![Screenshot-2017-11-20 Scratch 3 0 GUI(3).png](https://codetreeusercontent.s3.amazonaws.com/uploads/9dfb3f06-7fd0-4e72-b1d4-c32e7a366c9e/Screenshot-2017-11-20+Scratch+3+0+GUI%283%29.png)

When the green rectangle is selected with the reshape tool as shown, clicking on the left side of the ellipse where it overlaps the rectangle does not select the ellipse. Clicking on the part that does not overlap will select the ellipse.

fsih:
We have to think carefully about how this should work. This behavior was added intentionally because it was easy to accidentally change focus when working on the shape behind (for instance, what if you want to add a control point to the right edge of the rectangle right now? Should you have to move everything in front out of the way and then move it back after?)

chrisgarrity:
Yes, but I ran into this because I was editing the shape in front, but clicked off the edge when trying to add a point. In my case the front shape was entirely surrounded  by the back one, and I was unable to reselect the front shape with the reshape tool. I was trying to edit a very small shape on top of a full background.

Issue #150, opened 2017-10-27
Can't select shape where invisible handles are


jwzimmer-zz:
@fsih: 
> You can't select a shape where its invisible handles are. Makes it seem like select isn't working

PR #689, opened 2018-09-19
Use consistent hit testing

fsih:
### Resolves
Fixes https://github.com/LLK/scratch-paint/issues/150 (Can't select shape where invisible handles are)
Shapes should now be selectable where the invisible handles are.
Fixes https://github.com/LLK/scratch-paint/issues/520 (Hover outline doesn't necessarily match the next item you will select)
Mouse move now uses the same function as mouse down for hit detection, so these should match now
Fixes https://github.com/LLK/scratch-paint/issues/204 (Can't select front shape with the reshape tool if the back shape is selected)
Re-ordered the priority of hit detection so that you should be able to edit shapes that are fully in front of other shapes.

### Proposed Changes
1. Made mouse move and mouse down use the same hit algorithm in the reshape tool
2. Reordered priorities in the reshape tool so that handles that aren't visible don't interfere with hitting, and it's not impossible to edit shapes that are in front of other shapes

### Reason for Changes
Fix some frustrations with using reshape tool

### Testing
Test that hover selection (the thin blue line) is working as expected in select and reshape modes. Check that editing in reshape mode is working as expected.

https://llk.github.io/scratch-gui/handleBlock/

### DIFF
--- a/src/helper/hover.js
+++ b/src/helper/hover.js
@@ -30,6 +30,7 @@ const getHoveredItem = function (event, hitOptions, subselect) {
         }
     }
     const item = hitResult.item;
+    // If the hovered item is already selected, then there should be no hovered item.
     if (!item || item.selected) {
         return null;
     }
--- a/src/helper/hover.js
+++ b/src/helper/hover.js
@@ -2,7 +2,7 @@ import paper from '@scratch/paper';
 import {isBoundsItem, getRootItem} from './item';
 import {hoverBounds, hoverItem} from './guides';
 import {isGroupChild} from './group';
-import {sortHitResultsByZIndex} from './math';
+import {sortItemsByZIndex} from './math';
 
 /**
  * @param {!MouseEvent} event mouse event
@@ -25,7 +25,7 @@ const getHoveredItem = function (event, hitOptions, subselect) {
     // Get highest z-index result
     let hitResult;
     for (const result of hitResults) {
-        if (!hitResult || sortHitResultsByZIndex(hitResult, result) < 0) {
+        if (!hitResult || sortItemsByZIndex(hitResult.item, result.item) < 0) {
             hitResult = result;
         }
     }
--- a/src/helper/hover.js
+++ b/src/helper/hover.js
@@ -20,14 +20,14 @@ const getHoveredItem = function (event, hitOptions, subselect) {
     // sort items by z-index
     const items = [];
     for (const hitResult of hitResults) {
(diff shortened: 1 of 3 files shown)


src/helper/hover.js
```diff
@@ -12,22 +12,25 @@ import {sortItemsByZIndex} from './math';
...
+        if (!hitResult || sortItemsByZIndex(hitResult.item, result.item) < 0) {
+            hitResult = result;
         }
     }
-    items.sort(sortItemsByZIndex);
-
-    const item = items[items.length - 1];
-    if (!item) {
+    const item = hitResult.item;
+    if (!item || item.selected) {
```
ericrosenbaum, line 34: would be useful to have an explanatory comment here

src/helper/selection-tools/reshape-tool.js
```diff
@@ -211,11 +254,22 @@ class ReshapeTool extends paper.Tool {
...
     handleMouseMove (event) {
-        const hoveredItem = getHoveredItem(event, this.getHitOptions(), true /* subselect */);
+        const hitResult = this.getHitResult(event.point);
+        let hoveredItem;
+
+        if (hitResult) {
+            const item = hitResult.item;
+            if (item.selected) {
+                hoveredItem = null;
+            } else if (isBoundsItem(item)) {
```
ericrosenbaum, line 267: Looks like the reshape tool doesn't make use of grouping, so this if statement could be removed?

src/helper/selection-tools/reshape-tool.js
```diff
@@ -76,53 +77,84 @@ class ReshapeTool extends paper.Tool {
         paper.settings.handleSize = 8;
     }
     /**
-     * Returns the hit options to use when conducting hit tests.
-     * @param {boolean} preselectedOnly True if we should only return results that are already
-     *     selected.
+     * Returns the hit options for segments to use when conducting hit tests. Segments are only visible
+     * when the shape is selected. Segments take precedence, since they are always over curves and need
+     * to be grabbable. (Segments are the little curcles)
```
ericrosenbaum: typo :D

ericrosenbaum (dismissed):
Looks good, just a couple tiny nits and questions.

src/helper/selection-tools/reshape-tool.js
fsih: I realized this is relevant for reshape after all, since rasters and text are bounds items (although you can't really reshape either of them)

Issue #279, opened 2018-02-05
Stroke width does not get updated when stroke color is changed from none to color


chrisgarrity:
If you switch the stroke from none to a color I would expect the width to change from 0 to at least 1.

The version of the shape in the editor has no width (invisible unless selected), but the one on the stage is visible.

![stroke-issue](https://user-images.githubusercontent.com/399209/35820024-7674cbba-0a72-11e8-904a-9bfd33e68768.gif)

I would also expect the stroke width to automatically switch to 0 if you select no stroke.

carljbowman:
We should update to 1px when a color is selected after the stroke was transparent.
We should update to 0px when transparent is selected after the stroke had a color.

fsih:
@carljbowman what happens to the color when you have a transparent outline and a stroke width of 0, and then you increase the stroke width?

carljbowman:
We could set the stroke color to black, when the color is transparent and the stoke width is changed to anything larger than 0.

fsih:
@benjiwheeler suggests that if the stroke width is 0 and the stroke width gets increased, it should go back to whatever color it was before it was transparent, if any.

This could cause some surprises, like if you import an SVG with stroke widths 0, and increase the stroke width, you could get unexpected colors. But it could also be surprising to have a purple stroke, and have a black stroke after changing the stroke width.

I think it wouldn't be hard to implement this way fwiw

Edit: he added the issue at https://github.com/LLK/scratch-paint/issues/697

PR #695, opened 2018-09-20
Make stroke width 0 when transparent

fsih:
### Resolves
Fixes https://github.com/LLK/scratch-paint/issues/279

### Proposed Changes
When stroke color is transparent, make the width report 0
When stroke width is 0, make the color report transparent

When the stroke color is changed from transparent, make the width 1
When stroke width is changed from 0, make the color black

### Reason for Changes
Don't have hidden properties.
When you make changes, they are visible.

### DIFF
--- a/src/containers/stroke-color-indicator.jsx
+++ b/src/containers/stroke-color-indicator.jsx
@@ -3,13 +3,14 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import bindAll from 'lodash.bindall';
 import {changeStrokeColor} from '../reducers/stroke-color';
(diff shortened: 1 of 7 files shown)


benjiwheeler:
Code LGTM. Do you know why the checks are failing?

fsih:
Ah, oops! I need to fix the tests

fsih:
@benjiwheeler I got the stroke color to stay the same when you switch the width to 0 and then raise it up again, but it only works if you have something selected

fsih:
Fixed tests

benjiwheeler:
LGTM!

Issue #694, opened 2018-09-20
Adding a point in reshape doesn't create an undo state


fsih:
### Expected Behavior
Expect to be able to undo adding a point

### Actual Behavior
Adding a point doesn't add an undo state

### Steps to Reproduce
1. Go to reshape tool
2. Add lots of points to a shape
3. Notice that you can't undo, or they undo all at once, along with the last undoable action.

### Operating System and Browser
all

PR #696, opened 2018-09-20
Save after adding points

fsih:
### Resolves
https://github.com/LLK/scratch-paint/issues/694

### Proposed Changes
Make an undo state after adding a point

### Reason for Changes
If a point was added and not moved in reshape before, it didn't save the change


benjiwheeler (approved):
LGTM

Issue #557, opened 2018-07-15
Bitmap loads in the wrong place


fsih:
### Expected Behavior
Bitmap loads in the same place in 2.0 and 3.0

### Actual Behavior
Bitmap moves in this project
https://llk.github.io/scratch-gui/develop/#144885018

### Steps to Reproduce
Open https://llk.github.io/scratch-gui/develop/#144885018
Go to costumes tab
Make sure Reference sprite is selected
The bitmap should be centered, but it's offset down and to the right

### Operating System and Browser
Mac Chrome

thisandagain:
Looks like this is now working properly.

BryceLTaylor:
In this project the "cover" sprite does not appear correctly positioned in the paint editor: 
[3.0](https://llk.github.io/scratch-gui/develop/#247578672)
![screen shot 2018-09-26 at 10 22 52 am](https://user-images.githubusercontent.com/1088291/46086506-3a1bc400-c176-11e8-9bd4-bfe83cb1cf25.png)

[2.0](https://scratch.mit.edu/projects/247578672/#editor)
![screen shot 2018-09-26 at 10 23 01 am](https://user-images.githubusercontent.com/1088291/46086507-3c7e1e00-c176-11e8-816c-29fb1976f31e.png)

The elements in the costume were created as bitmaps in 2.0 (photos in this case) and then it was converted to vector so they could be positioned.   In 3.0, when it comes in as vector the images are positioned incorrectly.

You can reproduce this easily:
Create a new 2.0 project
Add a new sprite (should be in bitmap)
Draw a shape (circle, squiggle, anything) but keep it relatively small and near the center of the paint editor
Convert the sprite to vector
reposition the shape you draw closer to an edge of the editor
Save the project
Open the project in 3.0
Select your sprite and open the paint editor
Note: in 3.0 the image is drawn in the wrong place.  It will have moved farther from the center than you originally had it.  Sometimes they end up off the paint editor entirely.

BryceLTaylor:
These sprites also seem to be causing the project to slow down greatly when the paint editor is open to one of them.  This seems to be related to this issue: https://github.com/LLK/scratch-gui/issues/3172 and it causes [Geometry Dash](https://llk.github.io/scratch-gui/develop/#105500895)  to slow way down.

fsih:
https://github.com/LLK/scratch-paint/issues/632 seems to be another instance of this

fsih:
[...]

PR #755, opened 2018-10-30
Wait for other callbacks before load SVG code

fsih:
### Resolves
Fixes https://github.com/LLK/scratch-paint/issues/557

### Proposed Changes
This PR moves most of the code in the onLoad function for paper.project.importSvg into the window.setTimeout callback. (The diff isn't very helpful; it's mostly just code being moved)

### Reason for Changes
Bitmaps shouldn't load in the wrong place and "jump"
Imported rasters in their onLoad function set their own size and position, which they didn't have available previously. However the way the callbacks are set up, the individual rasters' onLoad functions aren't called until after the global SVG's onLoad function (even though the rasters' "loaded" status is "true" by the time the SVG's onLoad is called). This would cause rasters to appear and then reposition. Moving the code into the window.setTimeout causes the initializeSvg function to run after the rasters are finished initializing themselves. The item is only added to the paper view after all positioning is finished.

### Test Coverage
Tested loading projects 247578672 and 258120661

### DIFF
--- a/src/containers/paper-canvas.jsx
+++ b/src/containers/paper-canvas.jsx
@@ -27,6 +27,7 @@ class PaperCanvas extends React.Component {
         bindAll(this, [
             'setCanvas',
             'importSvg',
+            'initializeSvg',
             'maybeZoomToFit',
             'switchCostume'
         ]);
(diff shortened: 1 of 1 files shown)


fsih:
poke

paulkaplan (approved):
LGTM

Issue #742, opened 2018-10-25
Reshape tool doesn’t work to edit shapes that are on top of bitmaps in vector editor


fsih:
### Expected Behavior
Reshape tool should work

### Actual Behavior
Reshape tool ignores clicks when there is a raster behind the shape being reshaped

### Steps to Reproduce
1. Upload a bitmap image to the paint editor
2. Switch to vector mode
3. Draw a shape over the raster
4. Try to reshape it

### Operating System and Browser
Windows Chrome

fsih:
Console errors:
Uncaught TypeError: Cannot read property 'selected' of undefined
    at match (lib.min.js:205479)
    at filter (lib.min.js:135398)
    at Raster._hitTest (lib.min.js:135440)
    at Layer.hitTestChildren [as _hitTestChildren] (lib.min.js:135345)
    at Layer._hitTestChildren (lib.min.js:136269)
    at Layer._hitTest (lib.min.js:135438)
    at Project.hitTestChildren [as _hitTest] (lib.min.js:135345)
    at Project.hitTestAll (lib.min.js:135336)
    at ReshapeTool.getHitResult (lib.min.js:205561)
    at ReshapeTool.handleMouseDown (lib.min.js:205626)

PR #756, opened 2018-10-31
Fix issues handling rasters in reshape

fsih:
### Resolves
https://github.com/LLK/scratch-paint/issues/742

### Proposed Changes
Filter out rasters from several of the hit tests

### Reason for Changes
Rasters are capturing hits they shouldn't be capturing, causing reshape to fail when there is a raster in the background.

### Test Coverage
Tested using the reshape tool on a path in front of a raster. (To create a raster, e.g. upload a bitmap image, then convert to vector)


fsih:
poke

ericrosenbaum (approved):
LGTM!

PR #760, opened 2018-11-09
Revert "Wait for other callbacks before load SVG code"

paulkaplan:
Reverts LLK/scratch-paint#755

@fsih this PR actually causes a few different issues when integrated with the GUI.

1. It breaks the "zoom to fit" functionality
2. It introduces a strange issue where reordering costumes can cause the reordered costume to inherit the contents of the previous costume. See this gif

reorder-bug

I'm going to revert this PR so that we can deploy the GUI to beta without these fixes.

### DIFF
--- a/src/containers/paper-canvas.jsx
+++ b/src/containers/paper-canvas.jsx
@@ -27,7 +27,6 @@ class PaperCanvas extends React.Component {
         bindAll(this, [
             'setCanvas',
             'importSvg',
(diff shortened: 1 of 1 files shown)


fsih:
Sorry about the trouble, thanks for taking care of it Paul!

paulkaplan:
No trouble at all! I reviewed it, it was on me!

PR #762, opened 2018-11-11
Make mouse cursor react to what it's hovered over

towerofnix:
### Resolves

Resolves LLK/scratch-gui#2738. (/cc @benjiwheeler)

### Proposed Changes

When in select mode, change the cursor according to the handle that is currently being hovered over.

### Reason for Changes

In most paint/vector editors, the mouse cursor's image changes according to what handle the mouse is positioned over. For example, when over the top-right corner, a "resize in that direction" cursor is shown (an example of that cursor: a diagonal line with arrows on both ends)). This PR makes the Scratch 3.0 paint editor feel more like those familiar editors.

### Test Coverage

No automatic tests yet; only tested by hand. This is my first time working with Redux. While I'm hoping otherwise, some of the code might be kind of spaghetti!


src/reducers/cursor.js
```diff
@@ -0,0 +1,39 @@
+import log from '../log/log';
+
+const CHANGE_CURSOR = 'scratch-paint/cursor/CHANGE_CURSOR';
+const initialState = 'default';
```
fsih: It might be good to make constants for all of the possible values of the cursor string, so there'll be more safety than using strings.

src/helper/selection-tools/bounding-box-tool.js
```diff
@@ -35,10 +35,12 @@ class BoundingBoxTool {
      * @param {Modes} mode Paint editor mode
      * @param {function} setSelectedItems Callback to set the set of selected items in the Redux state
      * @param {function} clearSelectedItems Callback to clear the set of selected items in the Redux state
+     * @param {function} setCursor Callback to set the visible mouse cursor
      * @param {!function} onUpdateImage A callback to call when the image visibly changes
      * @param {?function} switchToTextTool A callback to call to switch to the text tool
      */
-    constructor (mode, setSelectedItems, clearSelectedItems, onUpdateImage, switchToTextTool) {
+    constructor (mode, setSelectedItems, clearSelectedItems, setCursor, onUpdateImage, switchToTextTool) {
```
fsih, line 43: The bounding box tool is used by many tools (e.g. oval tool, rect tool, text tool), so you'll have to update all those places, too.

src/helper/selection-tools/bounding-box-tool.js
```diff
@@ -105,27 +92,92 @@ class BoundingBoxTool {
...
     }
-    onMouseUp (event) {
+    onMouseUp (event, hitOptions) {
         if (event.event.button > 0 || !this.mode) return; // only first mouse button
         this._modeMap[this.mode].onMouseUp(event);
 
+        // Update the cursor to match the handle again. This has to be done a tick later because, if called
+        // immediately, the handles won't be visible yet (they are hidden while dragging).
+        setTimeout(() => {
+            this._updateCursor(event, hitOptions);
```
fsih: This doesn't feel like the right place for this code; it's kind of working around Redux's state machine and might have unintended consequences.

A simple alternative would be to change back to the default cursor on mouse up. Something more involved would be if all users of boundingBoxTool passed their getHitOptions functions to boundingBoxTool, so that boundingBoxTool could call updateCursor after it has reset the selection bounds, which is in onSelectionChanged.

src/helper/selection-tools/bounding-box-tool.js
```diff
@@ -74,29 +76,14 @@ class BoundingBoxTool {
      */
     onMouseDown (event, clone, multiselect, doubleClicked, hitOptions) {
         if (event.event.button > 0) return; // only first mouse button
-        const hitResults = paper.project.hitTestAll(event.point, hitOptions);
+        const {hitResult, hitResults, mode} = this._determineMode(event, multiselect, hitOptions);
```
fsih: You don't have to return hitResults anymore, since hitResult is sufficient to tell if anything was hit (which is what hitResults is used for below)

src/helper/selection-tools/bounding-box-tool.js
```diff
@@ -105,27 +92,92 @@ class BoundingBoxTool {
...
+    _updateCursor (event, hitOptions) {
+        const { mode, hitResult } = this._determineMode(event, false, hitOptions);
+        if (hitResult) {
+            if (mode === BoundingBoxModes.MOVE) {
+                this.setCursor('grab');
+            } else if (mode === BoundingBoxModes.ROTATE) {
+                this.setCursor('grab');
+            } else if (mode === BoundingBoxModes.SCALE) {
+                this.setSelectionBounds();
+                if (this.boundsPath.position.x === hitResult.item.position.x) {
```
fsih: If you wanted to simplify this part, another way to get the paper.Point of the corners is
boundsPath.topLeft, topRight, bottomRight, bottomLeft
There's also leftCenter, topCenter, rightCenter, bottomCenter
http://paperjs.org/reference/rectangle/

You could also add more helper data to the circles themselves if that would help; take a look at line 283.

fsih (changes requested):
Thanks! Overall the code design looks really good, and I think the feature will help a lot of people!

I found a couple issues testing it out.
- The grabbing cursor doesn't appear to change to the closed hand until after I mouse up:
cursor

- The cursor can get "stuck" if you mouse off the canvas while it's different. We should make sure that state gets cleaned up whenever we call deactivateTool (we can add that function to bounding box tool)
mouseoff

- Some of the other tools like oval and rectangle tool, and the bitmap tools, that use bounding box need to be updated.

Also there's some lint rules that need fixing in Travis: https://travis-ci.org/LLK/scratch-paint/builds/453455855?utm_source=github_status&utm_medium=notification

src/helper/selection-tools/bounding-box-tool.js
towerofnix: Good point, thanks. The problem I faced with this is that I need to do `handleMouseMove` on those tools as well, i.e. I need to add [this](https://github.com/LLK/scratch-paint/pull/762/files#diff-a4da56a05d313254413ded8e7d51b386R144) to the other tools. But those tools don't have access to `getHitOptions`. I'll probably have to go with a solution like you mentioned [here](https://github.com/LLK/scratch-paint/pull/762/#discussion_r234003522) (passing that function around). Input on how exactly that should be done, i.e. where that `getHitOptions` function should be stored (so we aren't duplicating code everywhere)? ATM I think we just define the `method(val) { dispatch(method(val)) }`-style thing for reducers in every file that uses those reducers, but obviously that's not an option for this. Is there a util function somewhere for this type of function? (`getHitOptions` works well as a separate function, BTW, since it doesn't need to be tied to any class - it doesn't ever reference `this`.)

EDIT: Actually the other classes *do* already have `getHitOptions`, maybe that will work.

towerofnix:
>The grabbing cursor doesn't appear to change to the closed hand until after I mouse up:

@fsih thanks for catching this (and for the review!). Just for reference, what browser are you using? I'm on Firefox/Linux.. I definitely forgot to test in Chromium, oops. Just want to make sure there's nothing particularly wonky going on in the case that we're using the same browser!

src/helper/selection-tools/bounding-box-tool.js
towerofnix: (Resolved because code is understood, still have to make commits for oval/text/(bitmap) tools.)

src/helper/selection-tools/bounding-box-tool.js
towerofnix: Yeah, I see. I think it would be better to take the latter option. My (long, unwieldy) paragraph above is somewhat relevant here:

> Good point, thanks. The problem I faced with this is that I need to do `handleMouseMove` on those tools as well, i.e. I need to add [this](https://github.com/LLK/scratch-paint/pull/762/files#diff-a4da56a05d313254413ded8e7d51b386R144) to the other tools. But those tools don't have access to `getHitOptions`. I'll probably have to go with a solution like you mentioned [here](https://github.com/LLK/scratch-paint/pull/762/#discussion_r234003522) (passing that function around). Input on how exactly that should be done, i.e. where that `getHitOptions` function should be stored (so we aren't duplicating code everywhere)? ATM I think we just define the `method(val) { dispatch(method(val)) }`-style thing for reducers in every file that uses those reducers, but obviously that's not an option for this. Is there a util function somewhere for this type of function? (`getHitOptions` works well as a separate function, BTW, since it doesn't need to be tied to any class - it doesn't ever reference `this`.)

NB - we do have a `getHitOptions` function in every relevant class (I think), i.e. the ones make use of a `boundingBoxTool`; but it feels like this function is getting duplicated around a lot. Not that it's particularly necessary to do this PR (I'm pretty sure, anyway), but could there be a better place to store that function, i.e. a utility function, considering the quoted paragraph above?

towerofnix:
> Some of the other tools like oval and rectangle tool, and the bitmap tools, that use bounding box need to be updated.

I added code to deal with this in all the appropraite places (based on a grep for `new BoundingBoxTool`).

> The cursor can get "stuck" if you mouse off the canvas while it's different. We should make sure that state gets cleaned up whenever we call deactivateTool (we can add that function to bounding box tool)

Should be fixed now (b6032ab313ad7b15d24928ced3297137dfa5a398).

> The grabbing cursor doesn't appear to change to the closed hand until after I mouse up:

[See above](https://github.com/LLK/scratch-paint/pull/762#issuecomment-439254828); I'll test on Chromium myself. Update - Weird. During *one* test, it was being wonky, like you described (sort of.. when I started dragging, it switched to the *normal* cursor, instead of the "grabbing" one; once I let go it switched back to the "can be grabbed" cursor). But after restarting Chromium (which, oops, for some reason nearly crashed my computer), it was working fine again (i.e. showed the correct cursors). Definitely strange.

Update - Besides figuring out why exactly it wasn't/isn't working on your computer, all your comments have been resolved (save for [a single thing about duplicating code less](https://github.com/LLK/scratch-paint/pull/762#discussion_r234072280), but that's not necessary as part of the PR). Thanks again for the super-thorough review!

src/reducers/cursor.js
towerofnix: Agreed. Resolved in 5e99ae3c5db57d8bfa6e52348e7b27bc29bca991 (I think that's the right place to put a constants file).

src/helper/selection-tools/bounding-box-tool.js
towerofnix: Thanks for the tip! Fixed in 2216ac22f79f061ecd071292d689efb02148cd14.

src/helper/selection-tools/bounding-box-tool.js
towerofnix: Sure, got it! Fixed in 217b5d71bd07bdbe36e2161cc70698f38008a088.

towerofnix:
Ooh, I found an extremely simple fix to [these shenanigans](https://github.com/LLK/scratch-paint/pull/762#discussion_r234003522). So the problem was that if we call `_updateCursor` *before* the bound handles are re-shown, it doesn't work (because the handles don't exist yet). To solve this, we previously used a `setTimeout()` so it would run on the next tick, by which point the bounds would be re-shown.

So.. here's the trick..

The function that shows those bounds, `setSelectionBounds`, is *already* in the `handleMouseUp` function (where we did the `setTimeout`). So fixing this is as simple as moving the call to `_updateCursor` after `setSelectionBounds`!

src/helper/selection-tools/bounding-box-tool.js
towerofnix: [Resolved](https://github.com/LLK/scratch-paint/pull/762#issuecomment-439273560) :)

fsih:
> Ooh, I found an extremely simple fix to these shenanigans.

Nice, it was there all along!
I'm seeing the mouse grab cursor issue on Windows 10, Chrome 70. I tried Firefox and Edge and didn't see the issue on those.

grab
Now that the cursor is a hand, I'm finding it a bit hard to click on small targets. For instance, here I was trying to grab the rotate handle. I think it would be good to leave the cursor as a pointer when you are hovering over shapes, especially since in addition to moving them, you can do stuff like multiselect shapes that doesn't make sense for a hand cursor.

src/reducers/cursor.js
```diff
@@ -0,0 +1,41 @@
...
+const initialState = Cursors.DEFAULT;
+
+const reducer = function (state, action) {
+    if (typeof state === 'undefined') state = initialState;
+    switch (action.type) {
+    case CHANGE_CURSOR:
+        if (typeof action.cursorString === 'undefined') {
+            log.warn(`Cursor should not be set to undefined. Use 'default'.`);
+            return state;
+        } else if (typeof action.cursorString !== 'string') {
```
fsih: You can use Object.values(Cursors).includes(action.cursorString) for type checking. We used to not be able to use Object.values, but I see it's being used in VM now so it should be fine

src/helper/selection-tools/bounding-box-tool.js
```diff
@@ -105,30 +93,85 @@ class BoundingBoxTool {
...
             this.removeBoundsPath();
         }
 
         return true;
     }
+    onMouseMove (event, hitOptions) {
+        this._updateCursor(event, hitOptions);
+    }
+    _updateCursor (event, hitOptions) {
+        const {mode, hitResult} = this._determineMode(event, false, hitOptions);
```
fsih, line 115: Let's only call setCursor if the cursor has changed, to avoid spamming actions.

fsih:
I figured out my delayed cursor changing issue! https://bugs.chromium.org/p/chromium/issues/detail?id=26723
Because chrome devtools takes focus, the mouse cursor doesn't update... but it's fine with devtools closed ᕕ( ᐛ )ᕗ

src/helper/selection-tools/bounding-box-tool.js
towerofnix: Agreed. Fixed in 5542e7e018e08c2ac6db1c75fd1e8d380164232b.

src/reducers/cursor.js
towerofnix: Right! Fixed in 9d0bd26f2f332b88f2754630abf241ae4dd28232.

towerofnix:
> Because chrome devtools takes focus, the mouse cursor doesn't update... but it's fine with devtools closed ᕕ( ᐛ )ᕗ

Phew, good that's figured out! The joys of web development.

> Now that the cursor is a hand, I'm finding it a bit hard to click on small targets. For instance, here I was trying to grab the rotate handle. I think it would be good to leave the cursor as a pointer when you are hovering over shapes, especially since in addition to moving them, you can do stuff like multiselect shapes that doesn't make sense for a hand cursor.

Good thinking! I definitely agree. I changed that in d27f1a6725aa0da95bbe8d019576a2834e5bc72d.

I also resolved the other two things you noted, so this is ready for re-review.

fsih (changes requested):
This looks great! Just one more thing, can we wait until onMouseDrag to set the cursor to "grabbing", just so it doesn't flash when you click?

towerofnix:
Thanks! Agreed, made the change (bb7bc155fdb79a128a8df42a0492b36e3a6bd4c4). Note I kept it so that the cursor changes to grabbing right away once you press down on the rotation handle; I think that feels intuitive in that case. And obviously you'll never somehow click on the rotation handle instead of the sprite, since the rotation handle appears only once you press down, and it's impossible for it to show up under where you pressed (AFAIK).

fsih:
Thanks! Yes, it makes sense for the rotate handle to change right away.

Oh no, I just discovered that this makes the eyedropper tool stop hiding the cursor ):
I'm not sure how the eyedropper tool is setting the cursor. I have to dig around...

Edit: here it is: https://github.com/LLK/scratch-paint/blob/badc549fdc5e45648382f55e8e109389cf7101fc/src/components/paint-editor/paint-editor.jsx#L203

fsih:
Darn, it's tricky. Eventually eyedropper probably needs to be refactored to use setCursor.

For now, I think a reasonable way to hack around it would be to have reducers/eye-dropper.js export ACTIVATE_EYE_DROPPER and DEACTIVATE_EYE_DROPPER, and reducers/cursor.js listen on those events and set the cursor to none and default respectively. We can file a bug to do the refactor later.

towerofnix:
>Darn, it's tricky. Eventually eyedropper probably needs to be refactored to use setCursor.

(Thanks for the investigation!) Practically speaking, would this be that difficult to do? It sounds like it might be easier to deal with it now than to work together a hack that definitely works and get rid of the hack later, once the eydropper code *is* updated. The cursor reducer file should be pretty easy to import into any file, not just the bounding box one, I think.

Edit: Okay, I took a look around... hypothetically the code would go in LoupeComponent, but I don't really see a good way of fitting it in there. Thoughts on what the "right" way to do it would be? ATM your reasonable hack is looking.. pretty reasonable! :P

fsih:
> Practically speaking, would this be that difficult to do?

I don't think the change would be too big, but the eyedropper is a hairy component that takes awhile to understand, and I don't want to make you dig into it for this unrelated change.

> Thoughts on what the "right" way to do it would be?

I might try getting rid of hideCursor and calling setCursor everywhere that activateEyedropper and deactivateEyedropper is being called currently, and see how that works. It'd be nicer if you didn't have to remember to always call both... but I don't have much time to think about it now unfortunately

towerofnix:
Thanks for the input! Eyedropper should be fixed in 5555527cd286b628cef66a900b834bd8fda213aa; that commit actually does a couple things. Per its description:

> * In the cursor reducer file, listen for the eyedropper being active or not, and appropriately adjust the cursor (hidden when active).
> * Remove the stuff related to the "hideCursor" property in ScrollableCanvas, keeping the important bits in a new "hideScrollbars" property.

fsih:
That sure is a lot of 5s

src/components/scrollable-canvas/scrollable-canvas.jsx
```diff
@@ -6,10 +6,7 @@ import styles from './scrollable-canvas.css';
 
 const ScrollableCanvasComponent = props => (
     <div
-        className={classNames(
-            props.style,
-            {[styles.hideCursor]: props.hideCursor}
-        )}
+        className={classNames(props.style)}
```
fsih: classNames is only needed to combine classes. Since there's only one, you can use it directly

src/components/paint-editor/paint-editor.jsx
```diff
@@ -200,7 +200,7 @@ const PaintEditorComponent = props => (
                 {/* Canvas */}
                 <ScrollableCanvas
                     canvas={props.canvas}
-                    hideCursor={props.isEyeDropping}
+                    isEyeDropping={props.isEyeDropping}
```
fsih: Just a really small thing, can we call this hideScrollbars? This way the ScrollableCanvas component doesn't imply that it knows about eyedropper or requires eyedropper to be used.

fsih (changes requested):
Thanks! The eye-dropper cursor looks good now aside from some very small nit-picks.

I noticed that occasionally, the N, E, S and W cursor is not the right shape for me. I think there's a very slight measuring issue, since it changes when I move the shape around.
nw

towerofnix:
Alright, pushed those tiny fixes (sorry for taking a little while to get to them).

>I noticed that occasionally, the N, E, S and W cursor is not the right shape for me. I think there's a very slight measuring issue, since it changes when I move the shape around.

Hmm.. should we use some sort of a threshold compare here, e.g. within 1 pixel or something along those lines? (We should be careful that doing that doesn't mess anything up when zoomed in.)

Also, I can't reproduce that, unfortunately (though I haven't tested in scratch-gui). So it's a little tough to tell if I've fixed it! :woman_shrugging:-:scream_cat:

fsih:
I looked into the wrong cursor issue, and logged hitResult.item.position.x and this.boundsPath.position.x. The difference is extremely small.
623.0985658452387 623.0985658452386
In paperjs when comparing floating point numbers, they consider 2 numbers a and b equal if abs(a - b) < 1e-8, so that would probably work here.

fsih (changes requested):
Can you remove package-lock.json?

fsih (commented):
Also the import classnames line can be removed:
https://github.com/LLK/scratch-paint/pull/762/files#diff-b2d9b36e6e6c50dad73d85ad3f578d1eR1

towerofnix:
Sure, fixed all the above! Looks like there's conflicts (probably with my other PR's changes), I'll take a peek.

towerofnix:
For some reason when I pull from develop and run, I get the error `_paintEditorMsgs2.default.messages is undefined; can't access its "en" property`, so I'm not too sure how to test a merge. (I get the same error even after a fresh git clone running from the `develop` branch.) I pushed a merge anyway, though, which does build with `npm start` (but I get that error when I open the scratch-paint playground).

fsih:
@towerofnix Sorry about that; another contributor already put up a fix for it though: https://github.com/LLK/scratch-paint/pull/785

fsih (approved):
It looks good to me!

Wow, it's the one month anniversary of this PR. Sorry we had you working on it for so long. But it looks great!

towerofnix:
>Wow, it's the one month anniversary of this PR. Sorry we had you working on it for so long. But it looks great!

Oh, cool! Happy anniversary PR. And no worries! It was actually really enjoyable and educational working on this, and your reviews were super helpful all along. It's pretty awesome to have completed this, though! :tada:

PR #774, opened 2018-11-16
Make circle (shift-oval) tool work for any direction dragged

### LINKED ISSUE #710: Shift-Circle only goes top-left/bottom-right
### Steps to Reproduce
1. Select the circle tool.
2. While holding the **shift** key, press and drag the mouse **to the top right.**

### Actual Behavior
The circle expands to the _bottom_ right.

### Expected Behavior
The circle expands to the top right.

Firefox 62, Ubuntu 18.04

### Additional information
- This issue is not present in Scratch 2.
- This issue is solved (in a subjectively less elegant manner) in the square tool.

towerofnix:
### Resolves

Fixes #710.

### Proposed Changes

Changed math used when shift key is pressed while creating an oval so that the circle can be drawn up-right and bottom-left of where the drag started.

### Reason for Changes

Bringing over features of Scratch 2.0. Note that dragging still isn't quite the same as in 2.0 (3.0 is based exclusively on the `x` offset instead of both); but with this PR, at least you *can* drag to the four relative quadrants rather than just two.

### Test Coverage

Tested manually. This math will need to be moved into a utility function, so that the bitmap oval tool works too. We might want to tweak the `diameter` variable to work however Scratch 2.0 does (try comparing the two versions yourself and you'll *feel* the difference); likewise, we might want to make the rectangle tool use this same math, so that rectangles scale the same way as ovals.

### DIFF
--- a/src/helper/math.js
+++ b/src/helper/math.js
@@ -131,6 +131,32 @@ const scaleWithStrokes = function (root, factor, pivot) {
     root.scale(factor, pivot);
 };
 
+// Get the size and position of a square, as in if the user were holding the shift key down while drawing the shape,
+// from the point where the drag started and the point where the mouse is currently positioned. (Note: This also works
+// for shapes like circles ("square ovals"), which fill the same dimensions.)
+const getSquareDimensions = function (startPos, eventPoint) {
+    // These variables are used for determining the relative quadrant that the shape will appear in.
+    // So if you drag up and right, it'll show up above and to the right of where you started dragging, etc.
+    let offsetX = eventPoint.x - startPos.x;
+    let offsetY = eventPoint.y - startPos.y;
+
+    // If the offset variables are zero, the shape ends up having zero width or height, which is bad.
+    // Deal with this by forcing them to be non-zero (we arbitrarily choose 1; any non-zero value would work).
+    offsetX = offsetX ? offsetX : 1;
+    offsetY = offsetY ? offsetY : 1;
+
+    const length = Math.abs(eventPoint.x - startPos.x);
+
+    const size = new paper.Point(
+        length * offsetX / Math.abs(offsetX),
+        length * offsetY / Math.abs(offsetY)
+    );
+
+    const position = startPos.add(size.multiply(0.5));
+
+    return {size, position};
+};
+
 export {
     HANDLE_RATIO,
     checkPointsClose,
@@ -138,6 +164,7 @@ export {
     expandBy,
     getRandomInt,
     getRandomBoolean,
+    getSquareDimensions,
     scaleWithStrokes,
     snapDeltaToAngle,
     sortItemsByZIndex
--- a/src/helper/math.js
+++ b/src/helper/math.js
@@ -145,7 +145,9 @@ const getSquareDimensions = function (startPos, eventPoint) {
     offsetX = offsetX ? offsetX : 1;
     offsetY = offsetY ? offsetY : 1;
 
-    const length = Math.abs(eventPoint.x - startPos.x);
+    // The length of the shape is the greater of the X and Y offsets.
+    const offsetDistance = eventPoint.subtract(startPos).abs();
+    const length = Math.max(offsetDistance.x, offsetDistance.y);
 
     const size = new paper.Point(
         length * offsetX / Math.abs(offsetX),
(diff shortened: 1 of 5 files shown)


src/helper/tools/oval-tool.js
```diff
@@ -89,7 +89,21 @@ class OvalTool extends paper.Tool {
         const downPoint = new paper.Point(event.downPoint.x, event.downPoint.y);
         const point = new paper.Point(event.point.x, event.point.y);
         if (event.modifiers.shift) {
-            this.oval.size = new paper.Point(event.downPoint.x - event.point.x, event.downPoint.x - event.point.x);
+            // These variables are used for determining the relative quadrant that the circle will appear in.
+            // So if you drag up and right, it'll show up above and to the right of where you started dragging, etc.
+            let offsetX = event.point.x - event.downPoint.x;
+            let offsetY = event.point.y - event.downPoint.y;
+
+            // If the offset variables are zero, the circle ends up having zero width or height, which is bad.
```
fsih: This step isn't necessary because if you look at line 128, if you draw an oval that's too small it will just get removed.

fsih (commented):
Thanks! This is definitely better than what we have now.

I agree that it would be nice to share this code with the bit oval tool. Code shared between tools is somewhat arbitrarily split up into helper js files right now. Maybe it would make sense in helper/math? You could also make a new helper file.

I also agree that it would be better if the behavior when holding shift worked more like it does in Scratch 2. I think it's making the diameter be the larger of offset x and offset y.

If you want to move on from this PR without adding those changes though that's fine, I still think we should merge it since it's incrementally better.

src/helper/tools/oval-tool.js
towerofnix: I think the problem was that if you drag a circle out to, for example, X = 100 and Y = 0 (relative to where you started dragging), you would *expect* to see a 100-unit circle, but there is none (because offsetX/Y = 0). The actual case where this is jarring is when you go from Y = 1 to Y = 0, because then the 100-unit (from X=100, Y=1) circle disappears.

towerofnix:
Alright, I split the math code into `helper/math.js` per your suggestion. This definitely helps clean up the code and makes it easier to tweak the math (e.g. to make it work like 2.0) in all places it's used.

In a similar vein of thought, I noticed I had to make a few identical changes to both `oval-tool.js` and `rect-tool.js` (which I implemented to make sure that it was useful across both tools). Maybe this is a little bit overkill, but I don't think so - Would it be useful to separate *all* the code which determines the size and position of a drawn shape into a shared math function? Ala `getShapeDimensions(startPos, eventPoint, modifiers)`; it would take into account alt, shift, and neither being pressed so that the other files don't have to have all that duplicate code themselves.

src/helper/math.js
```diff
@@ -131,13 +131,42 @@ const scaleWithStrokes = function (root, factor, pivot) {
     root.scale(factor, pivot);
 };
 
+// Get the size and position of a square, as in if the user were holding the shift key down while drawing the shape,
```
fsih: Could you switch this to a JS Doc, with
```
/**
 * @param
 * @returns
 */
```
? I know the rest of the code in this file doesn't have proper docs, but we can start here!

fsih (changes requested):
The new algorithm feels really good!

Do you mind switching  the bitmap oval and rect tools over to this algorithm too?

fsih:
> In a similar vein of thought, I noticed I had to make a few identical changes to both oval-tool.js and rect-tool.js

Yeah, those two tools are almost identical. I have been meaning to make them share a base class that they both extend (say ShapeTool). I'll probably refactor them if we add more shapes.

src/helper/math.js
towerofnix: Done (86c3bb1a4f4a1597a4fd04097dfbaa95b0758c89).

towerofnix:
> Do you mind switching the bitmap oval and rect tools over to this algorithm too?

Sure! Done, see 7e8bb8dfaf0b421060fe601cc778b512e609ce3d. Also JSDoc-ified that comment (86c3bb1a4f4a1597a4fd04097dfbaa95b0758c89).

fsih (approved):

fsih:
This looks great, thanks so much @towerofnix!

PR #785, opened 2018-12-09
Get proper paintMessages object properties

### LINKED ISSUE #784: Paint editor broken - "Cannot read property 'en' of undefined"
### Expected Behavior

It should load properly

### Actual Behavior

Error - Cannot read property 'en' of undefined

### Steps to Reproduce
 Go to https://llk.github.io/scratch-paint/

Mac OS, Google Chrome

epicfaace:
### Resolves

Resolves #784 (basically, https://llk.github.io/scratch-paint/ is currently broken)

### Proposed Changes

Change the paintMessages object properties so it doesn't give an error - it looks like #779 introduced this bug when updating to a new version of scratch-l10n

### DIFF
--- a/src/playground/reducers/intl.js
+++ b/src/playground/reducers/intl.js
@@ -14,13 +14,13 @@ const intlInitialState = {
     intl: {
         defaultLocale: 'en',
         locale: 'en',
-        messages: paintMessages.messages.en.messages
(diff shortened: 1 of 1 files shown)


chrisgarrity (approved):
Thanks for catching and fixing this!

PR #787, opened 2018-12-09
Add "testURL" config to fix test issues

epicfaace:
### Resolves

Tests failing -  #786

### Proposed Changes

Add "testURL" to jest config.

Issue #607, opened 2018-08-19
See exact middle of object when selected


DeleteThisAcount:
### Expected Behavior

Should show a circle in the exact center of selected object:

![](https://i.gyazo.com/6972b4b4ff857e8c13ef2973ead6d0d3.png)

### Actual Behavior

Does not:

![](https://i.gyazo.com/6c13c2deac447ae02299450cfdd4c203.png)

### Steps to Reproduce
Open paint editor
Select object
Notice missing circle

chexbox:
Related to #528?

DeleteThisAcount:
this is still relevant, and yes related to #528, not sure how active this repo is though

DeleteThisAcount:
This was added, so I'm going to close this.

Issue #528, opened 2018-06-26
Ability to move the rotation center of a group or shape


fsih:
We should decide how to surface this and how to save it

Some designs
![image](https://user-images.githubusercontent.com/2855464/49759584-23264180-fc90-11e8-80a2-fa6bd6e18be3.png)

benjiwheeler:
To clarify, this refers to a shape or group, NOT to the sprite as a whole.

That means this is a completely different issue from https://github.com/LLK/scratch-paint/issues/845

PR #788, opened 2018-12-10
Add a center marker to the selection bounding box

epicfaace:
### Resolves

Solves #607, part of #528. Should be merged after #785 and #787 .

### Proposed Changes

Add a center marker to the selection bounding box

image

### Reason for Changes

Helps with image design/editing, this feature was present in Scratch 2.0

### Test Coverage

No tests added for this


fsih (commented):
Cool, thanks for making this! It's something I had really wanted to get in but didn't have time to get to.

Could you remove the package-lock file and change to package.json from this PR?

fsih:
I talked to our designer about this change, and we don't want to be putting in visual changes this close to launch, since we don't have time to test for usability and bugs. He has concerns that the center dot looks like something you can interact with, and looks crowded when selecting small shapes. So I'm going to push this back to look at again in January. Thanks for working on this though!

DeleteThisAcount:
@fsih

I understand why the scratch team doesn't want to add something like this so close to launch.

Unfortunately I simply cannot make shapes without an exact center. 

If I want to make a sprite rotate, and it is even a few dozen pixels off it will look very odd and it will pivot on the actual middle.

Issue #768, opened 2018-11-13
Should we try to cap the stroke lower?


fsih:
Should we try to cap the stroke lower? Weird things happen if you go over 3 digit values… @kathymakes 
![image15](https://user-images.githubusercontent.com/1088291/48020513-1304ca80-e104-11e8-87c2-bcfb53024116.gif)

thisandagain:
@fsih @carljbowman @kathymakes Seems like capping at 100 would make sense.

fsih:
Some weird stuff can still happen at 100, but well, I guess it's less weird.
![image](https://user-images.githubusercontent.com/2855464/49965274-bf4b8500-feea-11e8-8ebe-e9e6d761319a.png)
If you look at the stage, for some reason that's how SVG renders really thick strokes...

PR #796, opened 2018-12-13
Limit stroke width to 100

fsih:
### Resolves
Fixes https://github.com/LLK/scratch-paint/issues/768

### Proposed Changes
Change the max stroke width to 100
I also bumped up the size of the stroke input so that it doesn't show '...' when you type in 100

### Reason for Changes
Strokes do really weird things when they're thick

### Test Coverage


benjiwheeler (approved):
Tested, LGTM!

PR #819, opened 2019-01-14
Make eyedropper work on touch, fix related bugs

paulkaplan:
Fixes an issue where you could not use the paint editor eye dropper on touch.

This was done by moving the eyedropper end callback from using mousedown to using mouseup, and make mousedown do the same thing as mousemove (i.e. hover). 

The effect of this is that touch devices can touch down, see the color, move and see the color change, then mouseup to finish. The non-touch experience is effectively the same, you can still hover to see colors and click to submit.

Existing hover action has not changed
hover-loupe

But touch experience works better
drag-loupe

Also fixed an  uncovered but with the loupe where it wouldn't draw the first time. Also includes a fix in an eyedropper-related line of code that was causing crashes reported from Sentry, where the code was assuming `previousTool` was not null.

### DIFF
--- a/src/components/loupe/loupe.jsx
+++ b/src/components/loupe/loupe.jsx
@@ -66,6 +66,11 @@ class LoupeComponent extends React.Component {
     }
     setCanvas (element) {
         this.canvas = element;
+        // Make sure to draw a frame when this component is first mounted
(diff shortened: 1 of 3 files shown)


rschamp (approved):
Makes sense, thanks @paulkaplan!

PR #826, opened 2019-02-08
End check for bounds at 0,0 instead of imageData.width, imageData.height

ktbee:
### Resolves

https://github.com/LLK/scratch-gui/issues/4409

### Proposed Changes

Invert the while loop order when checking for the bounds of a bitmap. This way we end our checks with an `x` and `y` value of 1. 

### Reason for Changes

Ending our bounds check at 0,0 will [pass 1,1 as the x and y values to calculate the skin's rotation center.](https://github.com/LLK/scratch-paint/blob/develop/src/hocs/update-image-hoc.jsx#L95-L96) Previously when the bitmap was empty, `getHitBounds` would get to the very bottom right corner of the imageData without finding a bound and would pass the bottom right coordinates (on my computer they were 960 and 720) as the x, y values for determining the rotation center, which [assumes the bound is at the top left of the image instead of the bottom right when it gets the image data.](https://github.com/paperjs/paper.js/blob/develop/src/item/Raster.js#L681)

By passing 1,1 as our x, y values for bitmaps without bounds, we are setting the rotation center as if our image's top left bound started near the top left of the stage, which sets the rotation center to be near the center of the stage, and allows the pen tool to draw on all parts of the stage. 

One thing that is tricky about this fix is that it only will fix the rotation center for existing projects if `getHitBounds` is run again. You can see this issue in [this project.](https://scratch.mit.edu/projects/284879849/) If you have the fix running locally and load this project, the pen tool will stay in the top left quadrant of the stage until you toggle between bitmap and vector for the empty bitmap skin. This isn't a problem for new projects.

### Test Coverage

I have an idea for how to test this, but I need to be able to create canvas images during the test. I tried `jest-canvas-mock` to mock canvas for the test, but it doesn't mock the image creating capability, so `getHitBounds` always finds the bitmap empty. I'm looking for a different option, but let me know if anyone has thoughts!

### DIFF
--- a/src/helper/bitmap.js
+++ b/src/helper/bitmap.js
@@ -329,10 +329,10 @@ const getHitBounds = function (raster) {
     let left = 0;
     let right = imageData.width;
 
-    while (top < bottom && rowBlank_(imageData, width, top)) ++top;
     while (bottom - 1 > top && rowBlank_(imageData, width, bottom - 1)) --bottom;
-    while (left < right && columnBlank_(imageData, width, left, top, bottom)) ++left;
+    while (top < bottom && rowBlank_(imageData, width, top)) ++top;
     while (right - 1 > left && columnBlank_(imageData, width, right - 1, top, bottom)) --right;
+    while (left < right && columnBlank_(imageData, width, left, top, bottom)) ++left;
 
     return new paper.Rectangle(left, top, right - left, bottom - top);
 };
--- a/src/helper/bitmap.js
+++ b/src/helper/bitmap.js
@@ -329,10 +329,10 @@ const getHitBounds = function (raster) {
     let left = 0;
     let right = imageData.width;
 
-    while (bottom - 1 > top && rowBlank_(imageData, width, bottom - 1)) --bottom;
     while (top < bottom && rowBlank_(imageData, width, top)) ++top;
-    while (right - 1 > left && columnBlank_(imageData, width, right - 1, top, bottom)) --right;
+    while (bottom - 1 > top && rowBlank_(imageData, width, bottom - 1)) --bottom;
     while (left < right && columnBlank_(imageData, width, left, top, bottom)) ++left;
+    while (right - 1 > left && columnBlank_(imageData, width, right - 1, top, bottom)) --right;
 
     return new paper.Rectangle(left, top, right - left, bottom - top);
 };
--- a/src/hocs/update-image-hoc.jsx
+++ b/src/hocs/update-image-hoc.jsx
@@ -89,9 +89,18 @@ const UpdateImageHOC = function (WrappedComponent) {
                 }
             }
             const rect = getHitBounds(plasteredRaster);
+            const imageData = plasteredRaster.getImageData(rect);
+
+            // If the bitmap has a zero width or height, save this information
+            // since zero isn't a valid value for on imageData objects' widths and heights.
+            if (rect.width === 0 || rect.height === 0) {
+                imageData.sourceWidth = rect.width;
+                imageData.sourceHeight = rect.height;
+            }
+
             this.props.onUpdateImage(
                 false /* isVector */,
-                plasteredRaster.getImageData(rect),
+                imageData,
                 (ART_BOARD_WIDTH / 2) - rect.x,
                 (ART_BOARD_HEIGHT / 2) - rect.y);
 


ktbee:
@kchadha had a good question when we were discussing this issue offline, and I wanted to answer it here since it ended up helping me find some unexpected behavior that I'm not sure if we should document anywhere. Previously I thought that this fix worked because the bitmap's height and width were being set to zero, which would allow this skin to use [the empty skin logic in scratch-render](https://github.com/LLK/scratch-render/blob/25df9f1ab7cc5187a14ad2ee6ab017b41f43b933/src/Skin.js#L114-L118) that SVGs and existing projects with 0x0 bitmaps use. @kchadha's question was, if changing the size of empty bitmaps is the fix, why can you use the pen tool on the whole stage with bitmaps that are the same size (the size of the stage), but not empty?

It turned out that the reason the fix worked had to do more with the different x and y positions (see the PR description) and not with the size of the bitmap. The empty bitmap's height and width [were changed to zero on this line](https://github.com/LLK/scratch-paint/blob/develop/src/hocs/update-image-hoc.jsx#L94) with this fix, but paper.js's `getImageData` function didn't behave in the way I was expecting. [It prevents us from having an empty 0x0 image in it's wrapper for canvas's `getImageData` function.](https://github.com/paperjs/paper.js/blob/6fc1b122d61d85605b4b498557b58fb6ca19be56/src/item/Raster.js#L679-L680) Instead if it finds the image is "empty", it uses a default size, that is the size of the full stage. 

The things I didn't expect about this:
- an empty bitmap skin doesn't use [the same logic](https://github.com/LLK/scratch-render/blob/4bf233ef361b6d7048d385b0c40bbdc19e4776e3/src/Skin.js#L108-L118) as the empty SVG or existing 0x0 bitmaps.
- An empty bitmap currently can't have a 0x0 size. Paper.js will make its size be full stage instead.
- With this fix, we pass the x, y coordinates as 1,1 because we increment `top` and `left` at the end of the while loop in `getHitBounds`, so once we reach 0,0, we then add one to the value we use for x and y. For a full-stage, non-empty bitmap, we pass 0,0 because the while loop's conditions are never true (the row is never blank), so we never increment `top` or `left`. ~~This means that non-empty full-stage bitmaps have a rotation center of 0,0 and empty full-stage bitmaps have a rotation center of -0.5, 0.5. I'm not sure what issues this would cause, but it felt worth noting.~~ (edit 2/26/19: I'm not sure where I was seeing these rotation center values now. Now I'm finding that `_rotationCenter` on the bitmap skin is (239.5, 179.5) for the empty bitmap and (240, 180) for the full stage bitmap)

I think this answers the original question and then some! Let me know if there are other ones or if I've misunderstood any of this. Also, where would be a good place to hold onto this sort of documentation? The compat wiki page?

cc @thisandagain

ktbee:
I've spent a fair amount of time figuring out testing for this PR, and it looks like it will take more work than I expected. @mzgoddard and I created a Chromeless testing set-up that integrated the scratch-paint React component and Jest, but I ran into another roadblock when it came to selecting and interacting with elements of the editor. So I've [pushed up my branch](https://github.com/ktbee/scratch-paint/tree/add-browser-canvas-tests) with the work so far, and will create an issue for continuing to work on this. If it makes sense to you all, perhaps we can hold off adding a test for this PR until we're able to interact with the paint editor in a browser test?

kchadha (commented):
@ktbee, is the reordering of the different checks being made also significant? If so, could you explain these changes?

ktbee:
@kchadha good Q, reordering the checks is how we finish checking the canvas in the top left corner at coordinates (0,0) instead of in the bottom right at coordinates that are the height and width of the stage ((960, 720) on my computer). This is because the variables `left`, `right`, `top`, `bottom` finish these checks with the values 1, 1, 1, 1 instead of 960, 960 , 720, 720. So we create a `paper.Rectangle` with a top left  X, Y value of (1,1) instead of (960, 720).

Creating a Rectangle with the top left bound for an empty bitmap in the top left corner of the stage sets the sprite's rotation center in the center of the stage, since paper.js's `getImageData` function sets the empty rectangle to be the size of the stage. Previously the top left bound of the empty bitmap was set to the bottom right of the stage, which made the rotation center set to be off stage.

kchadha:
@ktbee, I think the solution to this issue should result in the (x,y) of the top left corner at (0,0) instead of (1,1). There also seems to be some weird stuff going on with the rotation center...? From what I can tell the rotation center being reported for the empty bitmap sprite (with these changes) is (479, 359), probably directly related to the top left corner being at (1,1).

I'm right there with you about being confused about why this is not similar to the behavior of empty SVGs (esp. with the rotation center being at (0,0) for the empty svgs. It does seem like the behavior of these empty costumes (both bitmap and svg) is weird in a lot of different ways and there's probably a lot of different bugs at play here.

ktbee:
@kchadha now that I'm taking a look at this again, I'm not sure where or under what circumstances I was seeing that the rotation center was (0.5,0.5). I also see the (479, 359) values you mention when I console.log [the values passed for rotation center here](https://github.com/LLK/scratch-paint/blob/develop/src/hocs/update-image-hoc.jsx#L95-L96). So that at least is consistent!

I agree that it makes more sense to have the top left bound of the empty bitmap start at (0,0) instead of (1,1), so I will think of solutions around that. Another thought I had was to find a way to ensure a bitmap that doesn't have any hit bounds (ie is empty) keeps its width and height of zero. This would allow the empty bitmap to [behave in the same way empty SVGs do,](https://github.com/LLK/scratch-render/blob/25df9f1ab7cc5187a14ad2ee6ab017b41f43b933/src/Skin.js#L114-L118) which would cut down on even more inconsistencies. Though if you think that would be better tackled as a separate bug, I can file a separate issue.

kchadha:
> Another thought I had was to find a way to ensure a bitmap that doesn't have any hit bounds (ie is empty) keeps its width and height of zero. This would allow the empty bitmap to [behave in the same way empty SVGs do,](https://github.com/LLK/scratch-render/blob/25df9f1ab7cc5187a14ad2ee6ab017b41f43b933/src/Skin.js#L114-L118) which would cut down on even more inconsistencies.

This sounds great! 

(semi-unrelated: I'm still a little confused about why empty SVG's get their height and width marked as 2 x 2... but I definitely haven't investigated this issue at all)

> Though if you think that would be better tackled as a separate bug, I can file a separate issue.

I think I'll leave that part up to you based con how much of a rabbit hole this ends up being.

ktbee:
@kchadha I did some experimenting, and I believe my latest changes in this PR (plus [these changes in the VM](https://github.com/LLK/scratch-vm/compare/develop...ktbee:use-empty-bitmap-size?expand=1)) allow us to create 0x0 bitmaps when `getHitBounds` returns a Rectangle with a width or height of zero. To summarize the changes:

- I undid my changes to `getHitBounds`'s pixel check, since it wasn't necessary in this patch
- In scratch-paint: If `getHitBounds` returns zero for width or height, save that as the sourceWidth and sourceHeight on the bitmap's imageData object since imageData's width and height are read-only and can't be zero. 
- In scratch-vm in `updateBitmap`: if the imageData (now called bitmap) has a sourceWidth or sourceHeight of zero, use that as the canvas and costume's width and height. 0x0 canvases return a null blob, so do a check for blob before trying to read it as well.

This approach doesn't break any existing tests and I haven't found an issue with it yet from my manual tests. If this sounds good to you @kchadha, I'll squash these changes and create a PR for my scratch-vm changes as well.

ktbee:
Also I know that empty SVGs used to have a size of 2x2 in the UI (as you mentioned in your last comment), but as of now, their size is 0x0 even in the live site. Looks like some recent changes might have fixed that issue as well? 😅

cwillisf (approved):
I'm not sure I fully understand the implications here but the discussion+comments seem convincing. I think this is good to merge and we just need to test it well before release.
👍

PR #837, opened 2019-04-18
Add pencil tool

adroitwhiz:
This pull request adds a new vector-mode tool: the Pencil tool.

fig00

## What's it do?

The currently implemented Brush tool creates brush strokes that are outlined at draw-time:
fig01

While this approach has its advantages, including the ability to easily "cut through" brush strokes with the Eraser tool, making the brush more bitmap-like gives up a key advantage to using vector graphics in the first place: **reshapability**.

The Pencil tool creates brush strokes that are outlined via the "Outline" property, making them easily modifiable via the Reshape tool:
fig02

This is the way vector brush strokes were done in Scratch 2.0, which is one of the reasons I implemented it under the same name. One complaint resulting from the 2.0 -> 3.0 migration was that vector drawing was more cumbersome for those used to the 2.0 draw-reshape-fill workflow which enabled iterative improvements and flexibility.

## Details

This is what the user sees when they select the pencil tool:
fig03

The "Fill" is greyed out, and the "Outline" controls the pencil strokes' appearance. This communicates to Scratchers the difference between the pencil and brush tools: the brush gives you a filled shape, while the pencil gives you an outlined one. While the editor could automatically set the outline width to the brush width when the pencil tool is selected, I opted not to in order to avoid confusion.

There is also a "Smoothing" option. The higher it is set, the more the stroke's shape will be simplified. "Simplification" might be a better name, but also may be more confusing/opaque to the average Scratcher.

Simplification is done after the path is fully drawn and the mouse button is released. This might appear confusing, but I'm not sure of the performance implications of simplifying the path for each and every onMouseDrag event.

---

I've never done any React development before. While I tried my hardest to adhere to proper practices, there are probably some bugs in this code. Nevertheless, I hope this feature is robust and useful enough to make it into Scratch in some form or another.


thisandagain:
Thanks for investigating this! This is something we talked a lot about while designing the paint editor and ultimately decided to not include as we determined that understanding the difference between the two tools (paint brush vs. pencil) was difficult. In the future I recommend filing an issue to discuss features like this before spending the time to implement them.

/cc @carljbowman @fsih

PR #842, opened 2019-05-16
When converting to bitmap, call inlineSvgFonts() on SVG string instead of SVG element

### LINKED ISSUE #841: "Convert to Bitmap" doesn't embed SVG fonts
### Expected Behavior

Converting a vector sprite to a bitmap one should keep all the fonts intact.

### Actual Behavior

All fonts are replaced with the browser's "default" SVG font.

### Steps to Reproduce

1. Add some text to a vector costume
2. Click "Convert to Bitmap"

Linux, Firefox 66.0.2

adroitwhiz:
### Resolves

Resolves #841

### Proposed Changes

This changes the `convertToBitmap` function to pass the stringified SVG to `inlineSvgFonts()` instead of the SVG element.

### Reason for Changes

https://github.com/LLK/scratch-svg-renderer/pull/72 changed inlineSvgFonts() to operate on stringified SVGs, not SVG elements. It fails silently if given an SVG element--possibly something to look into changing as well?

### DIFF
--- a/src/helper/bitmap.js
+++ b/src/helper/bitmap.js
@@ -379,8 +379,9 @@ const convertToBitmap = function (clearSelectedItems, onUpdateImage) {
     // Get rid of anti-aliasing
     // @todo get crisp text https://github.com/LLK/scratch-paint/issues/508
     svg.setAttribute('shape-rendering', 'crispEdges');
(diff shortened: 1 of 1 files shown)


paulkaplan (approved):
Great, thanks for this!

PR #850, opened 2019-06-10
Add hex input

### LINKED ISSUE #69: Feature Request - Hex Codes
There should be a place where a user could input a hex code and the color (fill or outline whichever is selected) will change to that color.

towerofnix:
### Resolves

Resolves #69.

### Proposed Changes

1. Adds a hex input swatch, positioned between the transparent and eyedropper swatches. (https://github.com/LLK/scratch-paint/commit/afd933c872e664528f9820cf8510520a2363e918, https://github.com/LLK/scratch-paint/commit/47b50f7649d270e8a85a8929b716e64d4fd4b448)
    * It is exactly 7 characters wide, to make room for the full content of a `#123abc`-style string.
    * When clicked, it automatically selects the full content of the string.
    * Typing behavior is not well defined. It is meant to be used only for copying from and pasting to. (This is something that should be investigated, but I found messing around with the key-handling code to be rather a messy thing, especially whilst dealing with React.)
2. Changes the `<Slider>` element's code so that it does not rely on being a particular width. (https://github.com/LLK/scratch-paint/commit/bdd58e735d16293ef0eb2ae89c7b7651058ed9ba)
    * The width of the color picker is slightly greater and not strictly defined now, so the slider elements had to be adjusted to work at any width.
    * They still aim to be 150px+ but can be greater.
    * Constants were removed from the JS file (huzzah!), and the math was adjusted accordingly, offsetting the handle by a CSS-percentage of the container rather than an explicit number of pixels.
    * It controls exactly the same way it used to (stretched slightly). In order to keep the handle within the same range, an additional container had to be created, adding the 1/2-handle-width space on either end of the handle range.
3. Adds an additional state element to the color picker container: `hex`. (https://github.com/LLK/scratch-paint/commit/37eaf18395e01a2d8889e76605d2dc98acd4c315)
   * Although one would think this is redundant, it helps keep color precision tight - converting from hex to HSV and back tends to leave a small but noticeable change.

Edit: Screenshot below!

The full color picker, with the new hex input swatch, set to the color of the orange in the Scratch logo.

### Reason for Changes

All the technical reasons are described above. As for the feature itself, it is a much-requested feature on the Scratch forums - see [this topic](https://scratch.mit.edu/discuss/topic/349418/) for example; here's [another much older one](https://scratch.mit.edu/discuss/topic/195758/). Both of those gathered an impressive amount of 'support'.

### Test Coverage

No new tests implemented here (tested manually). Maybe a good idea? I've never approached component tests before, though.

### DIFF
--- a/src/components/color-picker/color-picker.css
+++ b/src/components/color-picker/color-picker.css
@@ -25,6 +25,10 @@
     justify-content: space-between;
 }
 
+.hex-input {
+    width: 7ch;
+}
+
 .row-header {
     font-family: "Helvetica Neue", Helvetica, sans-serif;
     font-size: 0.65rem;
@@ -59,13 +63,14 @@
 }
 
 .swatch {
-    width: 1.5rem;
-    height: 1.5rem;
+    width: 2rem;
+    height: 2rem;
     border: 1px solid #ddd;
     border-radius: 4px;
     box-sizing: content-box;
     display: flex;
     align-items: center;
+    justify-content: center;
 }
 
 .large-swatch-icon {
--- a/src/components/color-picker/color-picker.jsx
+++ b/src/components/color-picker/color-picker.jsx
@@ -284,7 +284,7 @@ class ColorPickerComponent extends React.Component {
                     <div className={classNames(styles.swatches, styles.hexSwatch)}>
                         <input
                             className={classNames(inputStyles.inputForm, styles.hexInput)}
-                            value={hsvToHex(this.props.hue, this.props.saturation, this.props.brightness)}
+                            value={this.props.hex}
                             onChange={this.handleHexChange}
                             onFocus={this.handleHexFocus}
                         />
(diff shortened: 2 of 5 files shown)


thisandagain:
Thanks! I really do appreciate the enthusiasm and support, but as per our contribution guidelines please don't open a PR for issues that don't have "help wanted" as a label. These types of features require discussion and often design from our team and we have limited bandwidth to review all of these different issues at the same time.

/cc @carljbowman

towerofnix:
@thisandagain Sure, sorry. That's ok. Is it alright if I ask about the ~10-15 PRs I've got open, in general much, much smaller than this one and on issues which are labelled help-wanted? I know you pinged various devs a while ago but it didn't seem to go anywhere, and they fix several bugs, a good few of which I feel are inherently worth working to resolve (LLK/scratch-blocks#1944 and LLK/scratch-gui#4147 for instance).

PR #854, opened 2019-06-13
Make scrollbars show up at initial zoom-to-fit size

### LINKED ISSUE #851: Scrollbars are not shown at initial zoom-to-fit scale
### Expected Behavior

_Please describe what should happen_

You should see scrollbars unless the editor is zoomed all the way out.

### Actual Behavior

Open the paint editor with the default project
See that you are zoomed-to-fit to the cat
See that you do not have scrollbars

paulkaplan:
Fixes https://github.com/LLK/scratch-paint/issues/851

### DIFF
--- a/src/containers/paper-canvas.jsx
+++ b/src/containers/paper-canvas.jsx
@@ -153,6 +153,7 @@ class PaperCanvas extends React.Component {
         } else if (this.shouldZoomToFit === true) {
             zoomToFit(isBitmapMode);
         }
+        this.props.updateViewBounds(paper.view.matrix);
         this.shouldZoomToFit = false;
     }
     importSvg (svg, rotationCenterX, rotationCenterY) {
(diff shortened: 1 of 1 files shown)


src/containers/paper-canvas.jsx
```diff
@@ -153,6 +153,7 @@ class PaperCanvas extends React.Component {
         } else if (this.shouldZoomToFit === true) {
             zoomToFit(isBitmapMode);
         }
+        this.props.updateViewBounds(paper.view.matrix);
```
fsih: Has this just always been a bug since launch? I'm surprised we missed that o_o

If this.shouldZoomToFit is false, this function should be a no-op. We only need to updateViewBounds if the conditionals above apply.

fsih (changes requested):

src/containers/paper-canvas.jsx
paulkaplan: Yeah I think it has been an issue since launch. 

And good call, I inlined the update in each individual case instead, see fc08bec

fsih (approved):
LG!

PR #856, opened 2019-06-20
Fix text input is delete to empty no update  image

slightc:
### Proposed Changes

Fix text input is delete to empty no update  image

### Reason for Changes

add update  image after clear textBox


thisandagain:
Please fill out the PR template available here:
https://github.com/LLK/scratch-paint/blob/develop/.github/PULL_REQUEST_TEMPLATE.md

PR #858, opened 2019-06-27
Fix dragging the whole page on iOS when using scrollbars

paulkaplan:
Fixes an issue where dragging scrollbars or the sliders on iOS would result in the whole page moving around underneath the dragger. This is because you cannot preventDefault on a `touchmove` unless you specifically mark it as non-passive. 

This is the current behavior (made on the iOS simulator)
dragging

Here is the new behavior
dragging-fixed


paulkaplan:
Another thing this causes is overscrolling the page on android tablets causes a page reflesh

fsih (approved):
LGTM! Sorry I missed this PR

PR #877, opened 2019-07-30
Update travis.yml to use jobs

fsih:
Clean up the yml file and switch to jobs


.travis.yml
```diff
@@ -1,44 +1,69 @@
 language: node_js
-sudo: required
-dist: trusty
 node_js:
 - 8
+- 10
```
rschamp: Since this is browser code, I don't think we actually care about testing this under more than one version of Node — probably sufficient just to update to 10 (and we should bump GUI to 10 as well).

.travis.yml
```diff
@@ -1,44 +1,69 @@
...
+      - provider: script
+        on:
+          branch: develop
+          condition: $TRAVIS_EVENT_TYPE == cron
+        skip_cleanup: true
+        script: npm run i18n:src && npm run i18n:push
+stages:
+- test
+- name: release
+  if: branch in (master, develop) and type != pull_request
```
rschamp, line 65: It looks like we also want to run this stage if the `type == cron`?

.travis.yml
```diff
@@ -1,44 +1,69 @@
...
-  skip_cleanup: true
-  script: npm run i18n:src && npm run i18n:push
+script: npm run $NPM_SCRIPT
+jobs:
+    include:
+    - env: NPM_SCRIPT=lint
+      node_js: 8
+    - env: NPM_SCRIPT=build
+      node_js: 8
+      if: not (branch in (master, develop) and type != pull_request)
```
rschamp, line 29: I think we might want to incorporate `type == cron` into this as well.

rschamp (changes requested):

.travis.yml
fsih: ray says ignore ray

.travis.yml
fsih: ray says ignore ray

.travis.yml
fsih: Changed to just node 8 to match gui

fsih:
Travis is now running deploy on a copy of this branch at https://travis-ci.org/LLK/scratch-paint/builds/567079102

Edit: looks like it worked

rschamp (approved):

PR #847, opened 2019-05-29
Display transparent center crosshair above items being dragged

adroitwhiz:
(This took ~15 minutes to proof out; I'm more interested in feedback/discussion than anything else)

### Resolves

Resolves #845

### Proposed Changes

This adds an additional guide layer containing a partially-transparent crosshair at the center of the canvas, which appears on top of sprites when they're being dragged:
Peek 2019-05-29 15-30

### Reason for Changes

This should make it much easier to center sprites.

Issue #845, opened 2019-05-24
Show center crosshair on top of artwork


benjiwheeler:
Feature request from Twitter, that I've heard asked for elsewhere:

“Alternate solution for allowing user to set a sprite's center: have the cross hair that shows the center to show on top of the custome in the editor. That way we can actually see the center. Pleeeeease!!!”

Maybe this could be shown only while moving a costume, so it's more clear that it's not a visible part of the costume itself?

This would help in centering costumes, without requiring users to just eyeball it.

BryceLTaylor:
This was recently addressed

PR #908, opened 2019-09-19
Center-Tool

### LINKED ISSUE #456: "Center" button when using rotation point tool
Building on #32, it would be handy to have a "center" button when the rotation point tool is active. This would make it so the rotation point is put in the middle of the costume.

Example usage - changing this...

A 5-point cartoon star; the rotation point crosshair is slightly offset from the middle A red ball, also with an offset rotation point

...to this:

The same star, but with the rotation point in the center The same ball, but with the rotation point in the center

(The costumes in the first set of images would rotate sort of wobbly; the second set would rotate around their actual middle, without any wobble. Compare rotating them in the 2.0 editor!)

kadhonn:
### Resolves

#456 #845 (and probably some more)

### Proposed Changes

I took the two commits from #847 and added a new "Center" button which automatically centers a figure by calculating the bounding boxes and centering it. Works for Bitmap and Vector mode.
I included a selfmade Icon for this, but I am no designer, so probably someone should redo it.

### Reason for Changes

There was still no replacement for the Scratch 2.0 center tool, which is dearly missed, so I tried to make my own.

### Test Coverage

Sorry, no tests for this right now


fsih:
Hopefully the tools we have added help to meet your needs, but let us know if you have feedback

kadhonn:
just checked it out, works really nice and does what i wanted, thanks :)

PR #909, opened 2019-09-19
Fix bitmap sprites opening with no default tool selected

### LINKED ISSUE #811: Paint Editor Tools should not be not selected by default
### Expected Behavior

No paint editor tools should not be selected by default 

### Actual Behavior

When clicking around on sprites while in the costumes tab the tool that's selected switches automatically creating a confusing experience because the options at the top of the paint editor change. 

### Steps to Reproduce
When opening an existing project (e.g Animate an Adventure Game) if you switch to "Costumes" tab no tools are selected for "Garnet". If you click on "Jake" no tool is selected and you see the flip options in the paint editor. If you then select "KO" from the sprites, the paintbrush tool is automatically selected. If you now switch to "Jake" the "paintbrush" is selected and you lose the options at the top.

fsih:
Fixes https://github.com/LLK/scratch-paint/issues/811

Set the tool correctly when the paint editor is opened on a bitmap sprite. The format starts off as null, and the mode starts off as SELECT

### DIFF
--- a/src/containers/paint-editor.jsx
+++ b/src/containers/paint-editor.jsx
@@ -24,7 +24,7 @@ import {resetZoom, zoomOnSelection} from '../helper/view';
 import EyeDropperTool from '../helper/tools/eye-dropper';
 
 import Modes from '../lib/modes';
-import {BitmapModes} from '../lib/modes';
+import {BitmapModes, VectorModes} from '../lib/modes';
 import Formats from '../lib/format';
 import {isBitmap, isVector} from '../lib/format';
 import bindAll from 'lodash.bindall';
@@ -77,7 +77,7 @@ class PaintEditor extends React.Component {
     constructor (props) {
         super(props);
         bindAll(this, [
-            'switchMode',
+            'switchModeForFormat',
             'onMouseDown',
             'onMouseUp',
             'setCanvas',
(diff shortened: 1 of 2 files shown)


src/containers/paint-editor.jsx
```diff
@@ -170,7 +171,8 @@ class PaintEditor extends React.Component {
                 log.error(`Mode not handled: ${this.props.mode}`);
                 this.props.changeMode(Modes.BRUSH);
             }
-        } else if (isBitmap(newFormat)) {
+        } else if (isBitmap(newFormat) &&
+                (this.props.mode in Modes && !(this.props.mode in BitmapModes))) {
```
kchadha: Why does this line of code not mirror the first `if` check above? E.g. above we only checked `if (isVector(newFormat) && (this.props.mode in BitmapModes))` as opposed to also adding `&& !(this.props.mode in Modes)` to that check as well...

I also wonder if we can consolidate these checks into helper functions (though not sure if there's a good way to do that... it seems like this is probably a good place for enums if `this.props.mode` can only ever be either `in BitmapModes` or `in Modes` but not both.

fsih: It's actually that Modes is a superset of BitmapModes. Maybe it would be good for understandability to make a VectorModes enum

fsih: I rewrote it adding VectorModes so hopefully it reads better now

fsih:
Hi @kchadha, what info do you need?

src/containers/paint-editor.jsx
kchadha: > It's actually that Modes is a superset of BitmapModes. Maybe it would be good for understandability to make a VectorModes enum

Ohhhh gotcha! I thought that Modes was vector modes. That makes sense now. Thank you for rewriting! I think the code is much clearer now! =]

kchadha (approved):
LGTM! Thank you for making those changes! I think the code is much clearer now.

PR #910, opened 2019-09-19
Fix use of eyedropper with color 2 of gradient

### LINKED ISSUE #861: Eyedropper does not work with secondary gradient color
When you have a gradient and have the "secondary" color selected (the one to the right), you can change it with the sliders. But if you try to use the eyedropper, it updates the color on the left instead of the selected color. 

eyedropper-bug

kchadha:
### Resolves

Resolves #861 

### Proposed Changes

Reorder code in `handleCloseFillColor` so that it does not reset the color index when the eyedropper is in use.

Previously, this code was correctly keeping the fill color menu open when the eye dropper was in use, but it was resetting the color index (the piece of redux state keeping track of which color in the gradient is currently selected) to `0` indicating that the first color in the gradient was selected. 

This was causing the bug in #861 where using the eyedropper on color2 of the gradient resulted in changing color1 instead.

### DIFF
--- a/src/containers/fill-color-indicator.jsx
+++ b/src/containers/fill-color-indicator.jsx
@@ -79,9 +79,14 @@ class FillColorIndicator extends React.Component {
         this.props.onChangeGradientType(gradientType);
     }
     handleCloseFillColor () {
-        if (!this.props.isEyeDropping) {
-            this.props.onCloseFillColor();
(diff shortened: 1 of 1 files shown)


fsih (approved):
LG, thanks! And thanks for the comments!

PR #919, opened 2019-11-08
Handle `null` colors in bitmap shape tools

### LINKED ISSUE #836: Undoing the creating a bitmap shape with a transparent fill give bsod in paint editor
### Expected Behavior

If you press the undo button the paint editor it should not throw an oops page.

### Actual Behavior

If you are in the bitmap editor and you draw a rectangle with a transparent fill, it appears black and if you undo it you get the blue oops screen.  In order to fill a bitmap with a transparent shade you have to use the eyedropper tool on the backdrop.

### Steps to Reproduce

Create a new project
Paint a new sprite
Click Convert to Bitmap
Select the square (or circle) tool
Select Fill
Choose the eyedropper
Click on the paint editor canvas where there is no shape
Draw a shape in the editor
Note: on the stage the it appears as a black filled in shape and does not appear in the editor.  The Shape is outlined
Click the shape again
Note: the shape appears in the editor as a black filled in shape
Click the undo button
Note: you get the blue oops screen.

editor

If you switch to the code tab and back the paint editor is back to normal and the shape is just a black shape.

Mac Chrome

adroitwhiz:
### Resolves

Resolves #836

### Proposed Changes

This PR:
1. Changes `RectTool.onSelectionChanged` and `OvalTool.onSelectionChanged` to set their `color` to `null` if the selected item's `color` is also `null`, instead of unconditionally calling the selected item's `color.toCSS` method (which will not exist if the color is `null`, causing an error to be thrown)
2. Changes `commitRectToBitmap` and `commitOvalToBitmap` to not draw anything if the `color` of the shape they're drawing is `null`

### Reason for Changes

1. fixes the crash when undoing the creation of a bitmap shape with no color
2. fixes such a shape appearing as black when clicked, whereas it should not appear at all

### DIFF
--- a/src/helper/bit-tools/oval-tool.js
+++ b/src/helper/bit-tools/oval-tool.js
@@ -76,7 +76,8 @@ class OvalTool extends paper.Tool {
                 this.thickness = this.oval.strokeWidth;
             }
             this.filled = this.oval.strokeWidth === 0;
-            this.color = this.filled ? this.oval.fillColor.toCSS() : this.oval.strokeColor.toCSS();
+            const color = this.filled ? this.oval.fillColor : this.oval.strokeColor;
+            this.color = color ? color.toCSS() : null;
         } else if (this.oval && this.oval.isInserted() && !this.oval.selected) {
             // Oval got deselected
             this.commitOval();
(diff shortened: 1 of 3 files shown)


fsih (approved):
LG, thanks!

fsih:
Hold off merging until next deploy

PR #920, opened 2019-11-08
Combine most `OvalTool` and `RectTool` functionality into `ShapeTool`

adroitwhiz:
This PR includes #919

### Proposed Changes

This PR moves code shared between `RectTool` and `OvalTool` into a `ShapeTool` base class that `RectTool` and `OvalTool` now inherit from. It does this for both bitmap and vector tools (there's both a bitmap `ShapeTool` and a vector `ShapeTool`).

The changed code lies in `handleMouseDrag`, which I modified to always set a positive value for the shape's size. This is to prevent rectangle shapes' `radius` attribute from getting messed up due to a paper.js bug.

### Reason for Changes

These changes greatly reduce the amount of duplicated code, making it easier to modify tool behavior in the future.

### DIFF
--- a/src/helper/bit-tools/shape-tool.js
+++ b/src/helper/bit-tools/shape-tool.js
@@ -1,0 +1,222 @@
+import paper from '@scratch/paper';
+import {getRaster} from '../layer';
+import {clearSelection} from '../selection';
+import {getSquareDimensions} from '../math';
+import BoundingBoxTool from '../selection-tools/bounding-box-tool';
+import NudgeTool from '../selection-tools/nudge-tool';
+
+/**
+ * Tool for drawing shapes (e.g. rectangles and ovals).
+ */
+class ShapeTool extends paper.Tool {
+    static get TOLERANCE () {
+        return 2;
+    }
+    /**
+     * @param {function} setSelectedItems Callback to set the set of selected items in the Redux state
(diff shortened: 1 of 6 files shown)


fsih (dismissed):
Thanks! A much-needed refactor!

fsih:
Hold for deploy

adroitwhiz:
A lot of this might have to be undone at a future point if rounded rectangles end up becoming a thing so I might leave this unmerged for a bit

src/helper/bit-tools/shape-tool.js
```diff
@@ -0,0 +1,222 @@
...
+        } else {
+            this.isBoundingBoxMode = false;
+            clearSelection(this.clearSelectedItems);
+            this.commitShape();
+
+            const shapeOptions = {
+                point: event.downPoint,
+                strokeScaling: false,
+                size: 0,
+                strokeJoin: 'round'
```
fsih, line 139: Now that I think about it, rectangles having rounded joins might be a bug. They were square in 2.0 and they're square in the vector editor. That would probably want to be in its own PR anyway if we wanted to fix that.

fsih (commented):
tiny shapes
When I test it now, I keep making tiny shapes when I click. I think stretchy paint may have made this issue easier to hit. We might want to bump up the tolerance, or have the tolerance depend on the zoom level.

PR #922, opened 2019-11-08
Skip processing of child paths in swapColorsInSelection

### LINKED ISSUE #567: Swapping colors in the gradient tool breaks on compound paths
### Expected Behavior

Swapping colors in the gradient picker should still continue to work regardless of shape.

### Actual Behavior

gradients

### Steps to Reproduce

1. Draw a shape with gradient 
2. Erase a hole into the shape to create a compound path
3. Swap colors in the gradient
4. Erase another hole into the the shape
5. Try swapping colors

Browser: Chrome Version 67.0.3396.99 (Official Build) (64-bit)
OS: macOS 10.13.6

adroitwhiz:
### Resolves

Resolves #567

### Proposed Changes

This PR changes the loop over all selected items in `swapColorsInSelection` to skip over child paths.

### Reason for Changes

The list of selected items contains both compound paths and their children.
Previously, the code would check if the path being iterated over was a child path, but if so it would instead swap the colors on the parent path. This is incorrect, as both the parent path and all of its child paths appear in the list of items and will eventually be iterated over.

This would fail on shapes with even numbers of holes because of the number of items being iterated over:

* For a single shape, there's one `Path` item, so swapping works
* For a shape with one hole, there's one `CompoundPath` item with two child `Path`s (one for the shape, one for the hole), so there are 3 total items and 3 total swaps and it works
* For a shape with two holes, there's one `CompoundPath` with three child `Path`s, so there are 4 swaps and it doesn't work

### DIFF
--- a/src/helper/style-path.js
+++ b/src/helper/style-path.js
@@ -6,6 +6,7 @@ import {getItems} from './selection';
 import GradientTypes from '../lib/gradient-types';
 import parseColor from 'parse-color';
 import {DEFAULT_COLOR} from '../reducers/fill-color';
+import {isCompoundPathChild} from '../helper/compound-path';
 
 const MIXED = 'scratch-paint/style-path/mixed';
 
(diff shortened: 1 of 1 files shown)


fsih (approved):
LG!

PR #925, opened 2019-11-14
Crosshair

fsih:
Add center crosshair overlay and center indicator to selection boxes


src/helper/selection-tools/bounding-box-tool.js
```diff
@@ -23,6 +25,7 @@ const BoundingBoxModes = keyMirror({
     ROTATE: null,
     MOVE: null
 });
+let anchorIcon;
```
paulkaplan, line 28: This global variable seems like a strange pattern. Why isn't it added as a property of the BoundingBoxTool prototype the same way things like `this.boundsPath` are?

src/helper/selection-tools/move-tool.js
```diff
@@ -101,7 +120,22 @@ class MoveTool {
...
         point.x = Math.max(0, Math.min(point.x, ART_BOARD_WIDTH));
         point.y = Math.max(0, Math.min(point.y, ART_BOARD_HEIGHT));
+        
         const dragVector = point.subtract(event.downPoint);
+        let snapVector;
+
+        // Snapping to align center. Only in select mode, because only select shows a center
+        // crosshair to line up.
+        const center = new paper.Point(ART_BOARD_WIDTH / 2, ART_BOARD_HEIGHT / 2);
+        if (!event.modifiers.shift && this.mode === Modes.SELECT) {
```
paulkaplan: Doesn't this collide with shift+drag being used for straight/diagonal movement?

src/helper/selection-tools/scale-tool.js
```diff
@@ -77,33 +78,36 @@ class ScaleTool {
...
                 this.lastSx = 1;
                 this.lastSy = 1;
             }
             this.pivot = this.origPivot;
         }
 
         this.corner = this.corner.add(delta);
-        const size = this.corner.subtract(this.pivot);
+        let size = this.corner.subtract(this.pivot);
+        if (event.modifiers.alt) {
```
paulkaplan: Hmmm this seems like an unrelated feature that it'd be good to put in separate PRs. Please try to keep PRs feature specific/small as possible to make it easier to review.

paulkaplan (changes requested):
I feel like the global `anchorIcon` comment is the only thing that needs addressing. But also I am a bit confused about the shift+drag behavior

fsih:
We finally got green light to put this in yesterday!

src/helper/selection-tools/move-tool.js
fsih: This is if NOT shift, so doesn't change shift+drag behavior

src/helper/selection-tools/bounding-box-tool.js
fsih: That was so that we would only need to load anchorIcon from SVG once! I fixed the code where it loads anchorIcon so it actually does that.

paulkaplan (approved):
@fsih code LG. One question I have about the functionality is it intended that the round center crosshair pops up above the costume layer but not above the selection highlight/bbox layer? It felt a bit strange, but not something that needs to change for this PR to go in, just wanted to mention it in case that wasn't the desired behavior.

fsih:
It's needed for you to be able to line up the selection crosshair over the black and white crosshair, but you're right that it looks weird that the rest of the blue lines overlap the center crosshair. Good point...

PR #932, opened 2019-12-09
Emit 1x1 blank bitmap data instead of full empty canvas

paulkaplan:
This fixes the issue where you cannot save blank bitmaps in scratch. We were trying to create PNG files which cannot have 0x0 size.

/cc @fsih


fsih (commented):
LG!

Bitmaps are "double resolution", so if we don't want the size to show up as 0x0, we might want to have the minimum exported size be 2x2 (which is 1x1 on the stage)?

paulkaplan:
It already shows up correctly as 1x1 on the tile.

PR #964, opened 2020-01-30
Alt+scale edge follows cursor

### LINKED ISSUE #963: Behavior when alt+scaling should be double the scale
### Expected Behavior
alt+scale would be more intuitive if the bound you were adjusting "stuck" to the mouse:
(This is a video of normal scaling, then holding alt for a bit, then normal scaling for a bit)
alt2

### Actual Behavior
Current alt+scale behavior:
alt

### Steps to Reproduce
In the paint editor, select something, then hold alt while resizing it

fsih:
Resolves https://github.com/LLK/scratch-paint/issues/963 (the gif of "Expected Behavior" is recorded from this implementation)


paulkaplan (approved):
Code LG, this is much better awesome!

PR #968, opened 2020-02-04
Fix Safari crash in color picker

### LINKED ISSUE #966: Color picker crashes on Safari
### Expected Behavior
Eyedropper should pick colors

### Actual Behavior
Eyedropper crashes
Edit: Seems to be browser-width related. This reproes for me at browser width 1257.

1. Go to costume tab (vector or bitmap, doesn't matter)
2. Stay on select tool
3. Open color picker (fill or outline, doesn't matter)
4. Click on eyedropper icon

### Steps to Reproduce
Click on eyedropper at the bottom of the color picker dropdown in the paint editor

Mac 10.14.6 Safari  13.0.5

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-paint/issues/966

### Proposed Changes

Multiply after subtracting, not before, to prevent rounding error. That rounding error (described further in https://github.com/LLK/scratch-paint/issues/966) caused the resulting imageData to be the wrong size. That in turn caused the `data.set` call to throw an error. 

I also put in a size check where that data is eventually used, since it is completely disconnected from where the data comes from so has no guarantee about size. Also added a log.warn in order to make it clear if this problem comes up again in another way.

### DIFF
--- a/src/components/loupe/loupe.jsx
+++ b/src/components/loupe/loupe.jsx
@@ -41,7 +41,16 @@ class LoupeComponent extends React.Component {
         const imageData = tmpCtx.createImageData(
             loupeDiameter, loupeDiameter
         );
-        imageData.data.set(this.props.colorInfo.data);
+
+        // Since the color info comes from elsewhere there is no guarantee
+        // about the size. Make sure it matches to prevent data.set from throwing.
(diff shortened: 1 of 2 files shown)


fsih (approved):
Thanks for investigating this paul!

Issue #937, opened 2020-01-06
Crash in stretchy-paint branch


fsih:
- Go to costume tab
- Click on sprite library
- Add amon
- Switch back to costume tab

paper-full.js:15038 Uncaught Error: TypeError: Cannot read property 'getSize' of null
    at onError (paper-full.js:15038)
    at XMLHttpRequest.onLoad (paper-full.js:15029)
    at XMLHttpRequest.xhr.onload (paper-full.js:13688)
layer.js:37 Uncaught TypeError: Cannot read property 'removeChildren' of undefined
    at clearRaster (layer.js:37)
    at _makeRasterLayer (layer.js:164)
    at setupLayers (layer.js:262)
    at PaperCanvas.componentDidMount (paper-canvas.jsx:62)
    at commitLifeCycles (react-dom.development.js:8770)
    at commitAllLifeCycles (react-dom.development.js:9946)
    at HTMLUnknownElement.callCallback (react-dom.development.js:542)
    at HTMLUnknownElement.dispatchEvent (<anonymous>)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:581)
    at invokeGuardedCallback (react-dom.development.js:438)
react-dom.development.js:9747 The above error occurred in the <PaperCanvas> component:
    in PaperCanvas
    in Connect(PaperCanvas)
    in div
    in ScrollableCanvasComponent
    in ScrollableCanvas
    in Connect(ScrollableCanvas)
    in div
    in div
    in div
    in PaintEditorComponent
    in InjectIntl(PaintEditorComponent)
    in PaintEditor
    in Connect(PaintEditor)
    in KeyboardShortcutsWrapper
    in Connect(KeyboardShortcutsWrapper)
    in CopyPasteWrapper
    in Connect(CopyPasteWrapper)
    in UndoWrapper
    in Connect(UndoWrapper)
    in SelectionComponent
    in Connect(SelectionComponent)
    in UpdateImageWrapper
    in Connect(UpdateImageWrapper) (created by PaintEditorWrapper)
    in PaintEditorWrapper (created by Connect(PaintEditorWrapper))
    in Connect(PaintEditorWrapper) (created by CostumeTab)
    in div (created by Box)
    in Box (created by AssetPanel)
    in div (created by Box)
    in Box (created by AssetPanel)
    in AssetPanel (created by CostumeTab)
    in CostumeTab (created by Connect(CostumeTab))
[...]

PR #974, opened 2020-02-08
Re-add centering

fsih:
Put back center snapping

This adds a fix for the crash, the repro steps were:
Switch to the costumes tab
File -> New
Switch to the costumes tab again

This also adds fading the crosshair when you're not dragging close to it, which I forgot to cherry-pick into the original centering PR, haha?

Demo is up at https://fsih.github.io/scratch-gui/centering/


adroitwhiz:
I'm not very familiar with React, so forgive my ignorance--is there any mechanism within React that can ensure the `PaperCanvas` element, which calls `paper.setup` and initializes the `paper.project.view`, "mounts" before all the other components?

Since it seems that moving `importSVG` calls to `onMouseDown` handlers is error-prone, this may indicate that https://github.com/LLK/scratch-paint/issues/937 isn't actually solved.

fsih:
@adroitwhiz the crash was caused by us calling paper's importSVG method, and then unmounting the paper-canvas component before the load callback could get called. That somehow leaves paper permanently in a broken state. We're getting around the issue here by avoiding calling importSVG at all, but we do call it in lots of other places, so you could probably hit this crash in other ways if you had really unlucky timing. We'll probably have to keep looking into it.

adroitwhiz:
It looks like [the importSVG method is synchronous if passed an SVG string](https://github.com/paperjs/paper.js/blob/develop/src/svg/SvgImport.js#L724), so if you import SVGs as strings via webpack's `raw-loader`, it should cause less weirdness (currently, SVGs are loaded as data URIs, which seem to follow [this code path](https://github.com/paperjs/paper.js/blob/develop/src/svg/SvgImport.js#L706)). It does seem rather fragile, though--maybe an "always synchronous" API (`importSVGString`?) could be added to Scratch's paper.js fork?

paulkaplan (approved):
LGTM, no crash and the crosshairs look the way ~they used to!~ they did in the previous version of this PR

PR #975, opened 2020-02-10
Fill space

### LINKED ISSUE #965: Fencing content to the workspace (Scrolling content phase 2)
Continuation of the work from https://github.com/LLK/scratch-paint/issues/166
Workspace resizes dynamically to fit the artwork (up to a point)

screen_shot_2020-01-30_at_1 51 23_pm

fsih:
### Resolves
Resolves https://github.com/LLK/scratch-paint/issues/965

### Proposed Changes
Feature announcement:
https://docs.google.com/document/d/122Phkf6ypP34tnONIwj00ltosifFxpRZM_8F62b25jY/edit

I broke out some commits that are independent from the main work.
- Commit 1 adds padding to the scrollbars to better match the specs, e.g. as seen here: https://projects.invisionapp.com/d/main?origin=v7#/console/18538067/385447939/preview?scrollOffset=209
- Commit 2 passes the mode through to more tools. Part of this work was to correct the mode being reported by several of the tools, which were being reported incorrectly before.
- Commit 3 adds a less-than-100% zoom level to the paint editor. This helps users take advantage of the extra space on smaller screens.

- Commit 4 makes the paint editor dynamically resize to fill the space available in the costume tab. Any time the canvas is resized, `paper.view.setViewSize` needs to be called for paper to redraw the elements on its canvas with the new canvas scale in mind, or the image will look stretched.

The code is handling 3 nested areas.
Screen Shot 2020-04-21 at 18 30 36
- The innermost, white checkered area is like the old paint editor. It matches the size of the stage, you can scroll to any part of it, and you can draw in it.
- The middle, blue checkered area is new in the vector editor. You can't scroll to it until you drag a part of your drawing onto it, at which point the scrollable area extends to include part of the blue checkered area. Then you can edit on it just like the white checkered area, until you drag your artwork off of it and scroll away, at which point you can't scroll back to it. The bitmap editor doesn't have this middle blue area, because stuff outside the stage sized area gets cropped in bitmap. The code related to switching how all the guide layers look when you switch between vector and bitmap is related to setting up and taking away this middle area.
- The dark blue outermost area is the wall that you usually hit when you try to drag things out of the paint editor right now. Mouse actions don't go past this wall. The difference is, you can see a sliver of it when you drag things to it. (You can see more if you have a huge screen, or zoom your browser way out.) The code related to masks is what's making vectors appear to go "under" this dark blue area.

Let me know if you have questions!

### Reason for Changes
More space to draw in the paint editor

The contrast between the checkers is reduced to reduce eye strain

The size of the checkered blue area in vector paint mode represents the maximum size a sprite can be drawn before setting its size to 100% will automatically clamp its size to a value less than 100%

### Test Coverage
Internal bug hunts

### DIFF
--- a/src/components/scrollable-canvas/scrollable-canvas.css
+++ b/src/components/scrollable-canvas/scrollable-canvas.css
@@ -1,5 +1,5 @@
 $scrollbar-size: 8px;
-$scrollbar-padding: 1px;
+$scrollbar-padding: 4px;
 
 .vertical-scrollbar, .horizontal-scrollbar {
     position: absolute;
@@ -10,17 +10,17 @@ $scrollbar-padding: 1px;
 .vertical-scrollbar-wrapper {
     position: absolute;
     width: $scrollbar-size;
-    right: 0;
+    right: $scrollbar-padding;
     top: $scrollbar-padding;
-    height: calc(100% - $scrollbar-size - $scrollbar-padding);
+    height: calc(100% - $scrollbar-size - 2 * $scrollbar-padding);
 }
 
 .horizontal-scrollbar-wrapper {
     position: absolute;
     height: $scrollbar-size;
     left: $scrollbar-padding;
-    bottom: 0;
-    width: calc(100% - $scrollbar-size - $scrollbar-padding);
+    bottom: $scrollbar-padding;
+    width: calc(100% - $scrollbar-size - 2 * $scrollbar-padding);
 }
 
 .vertical-scrollbar {
--- a/src/containers/paper-canvas.css
+++ b/src/containers/paper-canvas.css
@@ -1,6 +1,6 @@
 .paper-canvas {
-	top: 1px; /* leave room for the border*/
-	left: 1px;
+    top: 1px; /* leave room for the border */
+    left: 1px;
     width: calc(100% - 2px);
     height: calc(100% - 2px);
     margin: auto;
--- a/src/containers/paper-canvas.css
+++ b/src/containers/paper-canvas.css
@@ -1,7 +1,9 @@
 .paper-canvas {
-    width: 480px;
-    height: 360px;
+	top: 1px; /* leave room for the border*/
+	left: 1px;
+    width: calc(100% - 2px);
+    height: calc(100% - 2px);
     margin: auto;
     position: absolute;
-    background-color: #fff;
+    background-color: #D9E3F2;
 }
--- a/src/containers/paper-canvas.jsx
+++ b/src/containers/paper-canvas.jsx
@@ -228,7 +228,12 @@ class PaperCanvas extends React.Component {
                     window.setTimeout(() => {
                         // Detached
                         if (!paper.view) return;
-                        paper.view.setViewSize(paper.DomElement.getSize(paper.view.element));
+                        // Prevent blurriness caused if the "CSS size" of the element is a float--
+                        // setting canvas dimensions to floats floors them, but we need to round instead
+                        const elemSize = paper.DomElement.getSize(paper.view.element);
+                        elemSize.width = Math.round(elemSize.width);
+                        elemSize.height = Math.round(elemSize.height);
+                        paper.view.setViewSize(elemSize);
                         paperCanvas.props.updateViewBounds(paper.view.matrix);
                         paperCanvas.initializeSvg(item, rotationCenterX, rotationCenterY, viewBox);
                     }, 0);
--- a/src/containers/paper-canvas.jsx
+++ b/src/containers/paper-canvas.jsx
@@ -9,9 +9,10 @@ import log from '../log/log';
 import {performSnapshot} from '../helper/undo';
 import {undoSnapshot, clearUndoState} from '../reducers/undo';
 import {isGroup, ungroupItems} from '../helper/group';
-import {clearRaster, getRaster, setupLayers} from '../helper/layer';
+import {clearRaster, convertBackgroundGuideLayer, getRaster, setupLayers} from '../helper/layer';
 import {clearSelectedItems} from '../reducers/selected-items';
-import {ART_BOARD_WIDTH, ART_BOARD_HEIGHT, resetZoom, resizeCrosshair, zoomToFit} from '../helper/view';
+import {ART_BOARD_WIDTH, ART_BOARD_HEIGHT, CENTER, MAX_WORKSPACE_BOUNDS} from '../helper/view';
+import {clampViewBounds, resetZoom, setWorkspaceBounds, zoomToFit, resizeCrosshair} from '../helper/view';
 import {ensureClockwise, scaleWithStrokes} from '../helper/math';
 import {clearHoveredItem} from '../reducers/hover';
 import {clearPasteOffset} from '../reducers/clipboard';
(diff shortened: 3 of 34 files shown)


adroitwhiz:
It looks like there's still some blurriness caused by the canvas' CSS dimensions differing from its pixel dimensions:
image
image
image

src/containers/paper-canvas.jsx
```diff
@@ -253,9 +293,25 @@ class PaperCanvas extends React.Component {
...
+        // Fix incorrect paper canvas scale on browser zoom reset
+        this.recalibrateSize();
+        this.props.updateViewBounds(paper.view.matrix);
+    }
+    recalibrateSize () {
+        // Sets the size that Paper thinks the canvas is to the size the canvas element actually is.
+        // When these are out of sync, the mouse events in the paint editor don't line up correctly.
+        window.setTimeout(() => {
+            if (!paper.view) return;
+            paper.view.setViewSize(paper.DomElement.getSize(paper.view.element));
```
adroitwhiz, line 317:
```suggestion
            // Prevent blurriness caused if the "CSS size" of the element is a float--
            // setting canvas dimensions to floats floors them, but we need to round instead
            const elemSize = paper.DomElement.getSize(paper.view.element);
            elemSize.width = Math.round(elemSize.width);
            elemSize.height = Math.round(elemSize.height);
            paper.view.setViewSize(elemSize);
```

src/containers/paper-canvas.css
```diff
@@ -1,7 +1,9 @@
 .paper-canvas {
-    width: 480px;
-    height: 360px;
+	top: 1px; /* leave room for the border*/
+	left: 1px;
```
adroitwhiz: Nit: these appear to be tabs

```suggestion
    top: 1px; /* leave room for the border */
    left: 1px;
```

src/helper/layer.js
```diff
@@ -62,6 +63,20 @@ const getBackgroundGuideLayer = function () {
...
 };
 
+const _convertLayer = function (layer, format) {
+    if (isVector(format)) {
+        layer.vectorBackground.visible = true;
+        layer.bitmapBackground.visible = false;
+    } else if (isBitmap(format)) {
+        layer.bitmapBackground.visible = true;
+        layer.vectorBackground.visible = false;
+    }
```
adroitwhiz:
```suggestion
    layer.bitmapBackground.visible = isBitmap(format);
    layer.vectorBackground.visible = isVector(format);
```

src/containers/paper-canvas.jsx
```diff
@@ -266,10 +322,9 @@ class PaperCanvas extends React.Component {
         return (
             <canvas
                 className={styles.paperCanvas}
-                height="360px"
                 ref={this.setCanvas}
                 style={{cursor: this.props.cursor}}
-                width="480px"
+                resize="true"
```
adroitwhiz, line 332: Is `resize` a React/Paper-specific attribute?

src/helper/layer.js
```diff
@@ -241,30 +273,64 @@ const _makeDragCrosshairLayer = function () {
...
     const guideLayer = new paper.Layer();
     guideLayer.locked = true;
+    
+    const vWorkspaceBounds = new paper.Shape.Rectangle(MAX_WORKSPACE_BOUNDS);
+    vWorkspaceBounds.fillColor = '#ECF1F9';
+    vWorkspaceBounds.position = CENTER;
 
-    const vBackground = _makeBackgroundPaper(120, 90, '#E5E5E5');
-    vBackground.position = new paper.Point(ART_BOARD_WIDTH / 2, ART_BOARD_HEIGHT / 2);
+    const vBackground = _makeBackgroundPaper(180, 136, '#0062ff', 0.05);
```
adroitwhiz: Might be a good idea to avoid magic numbers here-- the background dimensions can be calculated from `ART_BOARD_WIDTH`, `ART_BOARD_HEIGHT`, `PADDING_PERCENT`, and  the checkerboard square size (8).

src/containers/paper-canvas.jsx
adroitwhiz: Not sure how well this works cross-browser-- it seems to work on Chrome and Firefox on Linux + Edge on Windows, but if it doesn't, some [additional styles](https://github.com/PullJosh/scratch-js/blob/master/scratch-js/Renderer.mjs#L315-L321) may need to be applied to the canvas.

adroitwhiz:
I'm also noticing some "jaggies" when I draw lines-- I drew a line at 100% zoom with a brush size of 1, and then zoomed in to see this result:
image
Current paint editor for comparison:
image

These don't appear to be related to the canvas blur-- even when rounding the canvas dimensions, lines sometimes still have "jaggies" when drawn.

src/components/scrollable-canvas/scrollable-canvas.css
```diff
@@ -10,17 +10,17 @@ $scrollbar-padding: 1px;
 .vertical-scrollbar-wrapper {
     position: absolute;
     width: $scrollbar-size;
-    right: 0;
+    right: $scrollbar-padding;
```
fsih, line 13: @adroitwhiz brings up this issue that it would be helpful if the padding counted as additional hitbox: https://github.com/LLK/scratch-paint/issues/1011

src/containers/paper-canvas.jsx
fsih: yes https://github.com/paperjs/paper.js/blob/a61e83edf2ed1870bd41bad135f4f6fc85b0f628/examples/css/style.css#L9

adroitwhiz: That looks like example CSS--is it actually included in this repo/imported from Paper somehow? I can't find any equivalent CSS here.

src/helper/layer.js
fsih: Good point

adroitwhiz:
It looks like the latest force-push ate some of my code review commits; not sure if that was intentional

src/containers/paper-canvas.jsx
fsih: Whoops! Here we go https://github.com/paperjs/paper.js/blob/9de1d4a2f36a2a4c62fbf71675bd92fb2b921460/src/view/View.js#L63

fsih:
Screen Shot 2020-05-06 at 23 01 08
@adroitwhiz I was able to get something similar on prod by moving the brush tool slowly at the outermost zoom level. I think the issue is that mouse events are coming in with integer coordinates. I don't think it's new, but it's worth filing

paulkaplan (dismissed):
Code looks ok to me and works as advertised. 

I might just be missing some context here, but I don't totally get why zoom jumps from .5 to .333 instead of being continuous all the way there. Again, I might just not understand the use case, but would like to know if you have a minute to explain. If there is a short version, a code comment in that zoom code I'm sure would help future us. That would be my only suggested change.

fsih:
Good catch @paulkaplan! I was thinking of the zoom + and - buttons and forgot to check the mouse scroll. It should be continuous up to .333 now.

paulkaplan (approved):

paulkaplan:
@fsih cool thanks

PR #981, opened 2020-02-12
Centering fade formula

fsih:
Change the formula for fading the center crosshair so that the crosshair is always opaque anywhere over the selection

Demo is up at https://fsih.github.io/scratch-gui/centering/

### DIFF
--- a/src/helper/selection-tools/bounding-box-tool.js
+++ b/src/helper/selection-tools/bounding-box-tool.js
@@ -212,15 +212,20 @@ class BoundingBoxTool {
             this.boundsRect.curves[4].divideAtTime(0.5);
             this.boundsRect.curves[6].divideAtTime(0.5);
             this.boundsPath.addChild(this.boundsRect);
-            
-            let anchorIcon = new paper.Group();
-            const vRect = new paper.Rectangle(new paper.Point(-1, -6), new paper.Size(2, 12));
-            const vRoundRect = new paper.Path.Rectangle(vRect, new paper.Size(1, 1));
-            const hRect = new paper.Rectangle(new paper.Point(-6, -1), new paper.Size(12, 2));
-            const hRoundRect = new paper.Path.Rectangle(hRect, new paper.Size(1, 1));
-            anchorIcon = vRoundRect.unite(hRoundRect);
-            vRoundRect.remove();
-            hRoundRect.remove();
+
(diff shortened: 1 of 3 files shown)


src/helper/selection-tools/bounding-box-tool.js
```diff
@@ -212,15 +212,20 @@ class BoundingBoxTool {
...
-            const vRoundRect = new paper.Path.Rectangle(vRect, new paper.Size(1, 1));
-            const hRect = new paper.Rectangle(new paper.Point(-6, -1), new paper.Size(12, 2));
-            const hRoundRect = new paper.Path.Rectangle(hRect, new paper.Size(1, 1));
-            anchorIcon = vRoundRect.unite(hRoundRect);
-            vRoundRect.remove();
-            hRoundRect.remove();
+
+            const vRect = new paper.Path.Rectangle({
+                point: [-1, -6],
+                size: [2, 12],
```
paulkaplan, line 218: Why these changes?

src/helper/selection-tools/move-tool.js
```diff
@@ -153,16 +158,43 @@ class MoveTool {
...
             // Show the center crosshair above the selected item while dragging.
             getDragCrosshairLayer().visible = true;
             this.firstDrag = false;
         }
-        const opacity = Math.max(0,
-            1 - ((CENTER.getDistance(this.selectionCenter.add(dragVector)) / CENTER.x) * (4 * paper.view.zoom)));
-        getDragCrosshairLayer().opacity = opacity;
+
+        let opacityMultiplier = 1;
+        const newCenter = this.selectionCenter.add(dragVector);
```
paulkaplan, line 178: Ooof this really needs some documentation explaining what the intention of this is.

paulkaplan (dismissed):
Code seems to work ok but I really think that opacity multiplier section needs a code comment describing its purpose. LG other than that.

src/helper/selection-tools/move-tool.js
fsih: Added comments

src/helper/selection-tools/bounding-box-tool.js
fsih: The code from before was leaving behind rectangles in the paper.view, which you could get in your selection by selecting all, and were weird immovable objects

PR #982, opened 2020-02-13
Fix not resizing the crosshair when switching tabs and coming back to the costume tab

fsih:
The bug was that when you opened a costume in the paint editor, zoomed in, switched tabs, switched back, and then dragged something to make the crosshair visible, the crosshair was super huge!

The case where you leave and come back to a costume you already opened is the case where we store the zoom level as a matrix. We were forgetting to readjust the crosshair size in that case.


kchadha (approved):
LGTM! Thank you for the fix!

PR #984, opened 2020-02-15
Set rotation center of empty SVGs to (0, 0)

### LINKED ISSUE #983: Empty vector costumes' rotation centers are incorrect
This looks like the *real* real cause of https://github.com/LLK/scratch-gui/issues/4233.

### Expected Behavior

When an empty vector costume is created, its rotation center should be (0, 0).

### Actual Behavior

When an empty vector costume is created, its rotation center is (240, 180)--this appears to be the vector version of #930.
This is due to `handleUpdateVector` [getting `drawnBounds`](https://github.com/LLK/scratch-paint/blob/develop/src/hocs/update-image-hoc.jsx#L117), which normally returns a rectangle around the active contents (with `x` and `y` set to the top-left corner of the contents). However, when there are no contents in the Paper layer, `drawnBounds`' `x` and `y` are (0, 0), which results in a rotation center of (240, 180).

### Steps to Reproduce

1. Delete all the contents of a vector costume
2. Download the project and extract the project.json
3. Observe the costume's rotation center

adroitwhiz:
### Resolves

Resolves #983

### Proposed Changes

This PR adds a special case in `handleUpdateVector` for when the `drawnBounds` of the active layer are 0x0.

### Reason for Changes

`bounds.x` and `bounds.y` are relative to the top left corner, but if there is no content in the active layer, they default to 0, making the "Scratch space" rotation center `((SVG_ART_BOARD_WIDTH / 2), (SVG_ART_BOARD_HEIGHT / 2))`, aka the upper left corner. This needs to be handled as a special case.

### Test Coverage

Tested manually

### DIFF
--- a/src/hocs/update-image-hoc.jsx
+++ b/src/hocs/update-image-hoc.jsx
@@ -134,6 +134,14 @@ const UpdateImageHOC = function (WrappedComponent) {
             scaleWithStrokes(paper.project.activeLayer, .5, new paper.Point());
 
             const bounds = paper.project.activeLayer.drawnBounds;
+
+            // `bounds.x` and `bounds.y` are relative to the top left corner,
+            // but if there is no content in the active layer, they default to 0,
+            // making the "Scratch space" rotation center ((SVG_ART_BOARD_WIDTH / 2), (SVG_ART_BOARD_HEIGHT / 2)),
+            // aka the upper left corner. Special-case this to be (0, 0), which is the center of the art board.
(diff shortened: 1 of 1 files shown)


fsih (dismissed):
LG, thanks!

fsih:
Hold off on merging until after this week's deploy

fsih (approved):

adroitwhiz:
Wait, are `SVG_ART_BOARD_WIDTH` and `SVG_ART_BOARD_HEIGHT` still the correct numbers to use after stretchy paint was merged? I was holding off on merging because I wanted to make sure.

fsih:
Oops sorry, I'll let you merge things

PR #991, opened 2020-03-27
Tighten bitmap selection bounds

### LINKED ISSUE #959: Bitmap selection should trim transparent pixels
### Expected Behavior
Click-drag selections in bitmap mode should probably trim transparent pixels off the edges of the selection. This will make them work better with the new center snapping. (Could there be some use case we're not thinking of where keeping the transparent pixels in the selection is useful?)

### Actual Behavior
Selection does not trim transparent pixels

### LINKED ISSUE #979: Selecting in bitmap should not include blank pixels in the selected area
### Expected Behavior

In bitmap mode, if you select an area around an image and there are many empty on the edge of the bounding box, the bounding box should shrink so that it just fits around the actual pixels.

### Actual Behavior

Currently it keeps the selected bounding box at the larger size.  With the addition of the centering tool, this will put the "center" of the shape in relation to the bounding box and not the selected shape.

### Steps to Reproduce

Create a project
Open the costume editor
Convert to Bitmap
Choose the selection tool
Select an area larger than the cat
Note: the bounding box for the selected area is whatever you had chosen and doesn't change shape to fit the non-transparent pixels in the selected area
Note: with the Centering tool the "center of the cat" will not be in the actual center of the cat if you selected a shape that is much bigger than the cat.

Mac Chrome

adroitwhiz:
### Resolves

Resolves #959
Resolves #979

### Proposed Changes

This PR tightens the areas selected by the bitmap selection tool to only include opaque pixels:
Peek 2020-03-27 14-14

As a supporting change, it modifies `getHitBounds` to optionally include a sub-rectangle of the passed raster as a parameter.

### Reason for Changes

See linked issues

### DIFF
--- a/src/helper/bitmap.js
+++ b/src/helper/bitmap.js
@@ -320,11 +320,18 @@ const columnBlank_ = function (imageData, width, x, top, bottom) {
     return true;
 };
 
-// Adapted from Tim Down's https://gist.github.com/timdown/021d9c8f2aabc7092df564996f5afbbf
-// Get bounds, trimming transparent pixels from edges.
-const getHitBounds = function (raster) {
-    const width = raster.width;
-    const imageData = raster.getImageData(raster.bounds);
+/**
+ * Get bounds around the contents of a raster, trimming transparent pixels from edges.
+ * Adapted from Tim Down's https://gist.github.com/timdown/021d9c8f2aabc7092df564996f5afbbf
+ * @param {paper.Raster} raster The raster to get the bounds around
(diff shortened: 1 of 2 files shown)


fsih (dismissed):
LG, thanks!

fsih (dismissed):
The merged code here is from adding the ability via stretchy-paint to create selection boxes that are entirely outside of the bounds of the raster paint area. I tested it out and it looks like it's working, but you might want to review the merged code

fsih (approved):

PR #992, opened 2020-03-29
Fix SVG element with invalid ID when importing

### LINKED ISSUE #53: "Names consisting only of numbers are not supported" error
`<g id="1" transform="translate(46.000000, 37.000000)" fill-rule="nonzero">`
SVGs with IDs consisting of only numbers can't be imported, such as one with the above substring.
That seems quite brittle. By default some vector applications and SVG optimizers give groups a number.
from @thisandagain

---
Moved from LLK/paper.js#13
/cc @fsih @thisandagain @rschamp @gnarf

apple502j:
### Resolves
Resolves LLK/scratch-svg-renderer#53
Resolves some parts of LLK/scratch-gui#4944

### Proposed Changes
Changes SVG elements' id to uid, if it is invalid(has only numbers) when importing.

### Reason for Changes
paper.js does not parse SVGs with numeric id.

### DIFF
--- a/src/containers/paper-canvas.jsx
+++ b/src/containers/paper-canvas.jsx
@@ -4,6 +4,7 @@ import React from 'react';
 import {connect} from 'react-redux';
 import paper from '@scratch/paper';
 import Formats from '../lib/format';
+import uid from '../lib/uid';
 import log from '../log/log';
 
 import {performSnapshot} from '../helper/undo';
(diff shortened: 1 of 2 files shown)


apple502j:
Gonna close this one, @adroitwhiz is doing a better job on  LLK/paper.js#34

PR #997, opened 2020-04-11
Adjust reshape mode handle tolerance

### LINKED ISSUE #691: In reshape tool, you often add a point when you mean to drag a point
### Expected Behavior
You should mostly do what you expect to do

### Actual Behavior
I add points instead of dragging them all the time >:S maybe you shouldn't be able to make overlapping points.
addpoint

### Steps to Reproduce
In the reshape tool, drag points around to make a shape

all

### LINKED ISSUE #737: Hit targets for Bezier curve handles feel small
Hit targets for the Bezier curve handles (diamonds) feel small. They are hard to click and drag. +1 @carljbowman

adroitwhiz:
### Resolves

Resolves #691
Resolves #737

### Proposed Changes

This PR changes the handle size used by the reshape tool to match the tool's `TOLERANCE` constant such that clicking anywhere on a handle (including the handle's stroke/outline) will count as clicking the handle.

It also increases `TOLERANCE` to 5 so that the now-dynamically-sized handles retain their previous size.

### Reason for Changes

This makes handles' clickable regions match their appearance on-screen, making it much easier to move points around.

### DIFF
--- a/src/helper/selection-tools/reshape-tool.js
+++ b/src/helper/selection-tools/reshape-tool.js
@@ -33,6 +33,18 @@ const ReshapeModes = keyMirror({
 class ReshapeTool extends paper.Tool {
     /** Distance within which mouse is considered to be hitting an item */
     static get TOLERANCE () {
+        return ReshapeTool.HANDLE_RADIUS + ReshapeTool.HANDLE_PADDING;
+    }
+    /**
+     * Units of padding around the visible handle area that will still register clicks as "touching the handle"
+     */
+    static get HANDLE_PADDING () {
+        return 1;
+    }
+    /**
+     * Handles' radius, including the stroke
+     */
+    static get HANDLE_RADIUS () {
         return 5.25;
     }
     /** Clicks registered within this amount of time are registered as double clicks */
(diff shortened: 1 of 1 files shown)


fsih:
Reshape points, before and after
before circle test
after circle test

Reshape handles, before and after
before diamond test
after diamond test

It looks better, but it still seems like clicking on the stroke of the handle/point doesn't register as a click on the handle/point

fsih (commented):
From https://github.com/LLK/paper.js/commit/2d790a3f34922647abb5c04b1ed929daf93fe539
- It looks like the line width is actually 2.5
- The diamond-shaped handles are tricky because the hit area is not a circle, so you can appear to click the corners without hitting the hit area. Maybe it's ok if we leave off the corners.

adroitwhiz:
Thanks for tracking down the stroke width! I've adjusted the calculation to match--it should feel slightly better now though I'm not sure by how much.

Should there be some additional "fudge factor"?

fsih (commented):
Let's change the tolerance to 6 (while keeping the handle size at 5.25*2). The mouse positions are snapping to the nearest integer, so I'm not sure if decimal values are doing anything.

Let's add a comment that the selection stroke width is 2.5 (maybe link to the code in paper) and that we're adding .75 "padding"

src/helper/selection-tools/reshape-tool.js
```diff
@@ -33,7 +33,14 @@ const ReshapeModes = keyMirror({
 class ReshapeTool extends paper.Tool {
     /** Distance within which mouse is considered to be hitting an item */
     static get TOLERANCE () {
-        return 4;
```
fsih, line 36: I think it would make more sense to define this as a HANDLE_SIZE of 8 with a TOLERANCE of 2.25. I think tolerance is usually defined as the space outside of the visible thing you're clicking which still counts as a click on the object.

fsih (commented):
1 last nit

fsih (approved):
LG, thanks!

PR #1000, opened 2020-04-13
Use divideAt in PointTool.addPoint so that adding new points doesn't affect curves' shape

### LINKED ISSUE #998: Adding a point to a curved segment with the reshape tool affects the segment's shape
### Expected Behavior

When you use the reshape tool to add a point in the middle of a curved segment, it should not affect that segment's shape. This is how Inkscape, for instance, behaves:
Peek 2020-04-11 11-05

### Actual Behavior

Adding new points to segments affects the shape of those segments:
Peek 2020-04-11 11-07

### Steps to Reproduce

1. Switch to the reshape tool
2. Select a shape that contains curved segments
3. Click along one of those curved segments to add a point
4. Observe how the segment's shape changes

All

adroitwhiz:
### Resolves

Resolves #998

### Proposed Changes

This PR replaces the subdivision logic in `PointTool.addPoint` with Paper.js' `divideAt` method.

### Reason for Changes

Currently, when adding points using the reshape tool, the shape of the curve segment where the point was added changes, because of how the `addPoint` subdivision logic works.

Paper.js' `divideAt` method will instead add a new point in the middle of a curve segment while not changing that segment's shape.

### DIFF
--- a/src/helper/selection-tools/point-tool.js
+++ b/src/helper/selection-tools/point-tool.js
@@ -1,4 +1,3 @@
-import paper from '@scratch/paper';
 import {HANDLE_RATIO, snapDeltaToAngle} from '../math';
 import {getActionBounds} from '../view';
 import {clearSelection, getSelectedLeafItems, getSelectedSegments} from '../selection';
@@ -59,7 +58,7 @@ class PointTool {
             }
             hitProperties.hitResult.segment.selected = true;
         }
-        
+
         this.selectedItems = getSelectedLeafItems();
     }
     /**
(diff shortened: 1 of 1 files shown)


fsih (commented):
Awesome! I didn't realize this existed. Works way better!

It handles straight edges differently. The old code would add a curved point, whereas the new one adds a point without handles. I can see people preferring either way. I like drawing a polygonal outline with the line tool, then adding in the curves with the add point tool. Maybe we should add this question to the design meeting as well?

fsih:
Let's make all added points curved for now (i.e. add handles to them), to keep the behavior similar to the current behavior.

adroitwhiz:
@fsih I replaced the custom smoothing logic with Paper.js' built-in `smooth` for segments added in the middle of straight lines. Is that alright?

src/helper/selection-tools/point-tool.js
```diff
@@ -68,51 +67,16 @@ class PointTool {
...
-            handleIn = handleIn.multiply(3 * handleOut.length / handleIn.length);
-        }
-        if (handleOut.length > 3 * handleIn.length) {
-            handleOut = handleOut.multiply(3 * handleIn.length / handleOut.length);
-        }
+        const newSegment = hitProperties.hitResult.item.divideAt(hitProperties.hitResult.location);
 
-        const beforeSegment = hitProperties.hitResult.item.segments[hitProperties.hitResult.location.index];
-        const afterSegment = hitProperties.hitResult.item.segments[hitProperties.hitResult.location.index + 1];
+        // If we're adding a point in the middle of a straight line, it won't be smooth by default, so smoothe it
```
fsih: Typo smooth

fsih (dismissed):
LG! Might want to wait until the week after the paper merge goes in to merge this.

fsih (approved):
LG, thanks!

Let's wait on this one, too

PR #1006, opened 2020-04-20
Improve broad brush handling

### LINKED ISSUE #320: End caps of thick strokes sometimes not rounded
Expected: Ends of strokes are always rounded.

Actual: Sometimes the end caps are sharp edged:

paint-arrow

I'm also seeing a warning in the console (not sure if it's related): `Skipping a merge.`

### LINKED ISSUE #715: Vector brush strokes look bad
### Expected Behavior
They should look similar to bitmap brush strokes

### Actual Behavior
Almost every time I use the vector brush, it doesn't look smooth
image
There are nicks whenever you change directions, and caps can look detached or sharp instead of rounded.

### Steps to Reproduce
Draw a wiggly line with the vector brush tool

Windows 10 Chrome

adroitwhiz:
### Resolves

Resolves #715
Resolves #320

### Proposed Changes

- Clear `BroadBrushHelper.lastVec` in `onBroadMouseDown`
- Don't add both `event.middlePoint` and `event.point` to the brush path in `onBroadMouseMove`
- Set the brush path angle to the proper vertex normal
- Use the correct mouse delta in `onBroadMouseUp`

### Reason for Changes

Taken together, these changes significantly improve the quality of brush strokes created by the broad brush algorithm:

- Clear `BroadBrushHelper.lastVec` in `onBroadMouseDown`
  - `onBroadMouseMove` checks for the existence of `lastVec` to determine whether to add end caps. If `lastVec` is not cleared, it will hold the value from the previous brush stroke, causing an erroneous end cap to appear at an incorrect angle at the beginning of every brush stroke after the first.

- Don't add both `event.middlePoint` and `event.point` to the brush path in `onBroadMouseMove`
  - This appears to be the cause of most of the "wrinkles" in the brush stroke:
  - image
  - `event.middlePoint` is just a linear interpolation between the last event point and `event.point`, so it doesn't need to be added to the brush path.

- Set the brush path angle to the proper vertex normal
  - This is hard to explain. Previously, the brush's points were expanded outwards at the angle of the mouse delta event before them, like so:
  - image
  - Now they take the average angle of the deltas before and after them:
  - image
  - This also makes the brush edges less jagged.

- Use the correct mouse delta in `onBroadMouseUp`
  - The mouse up event's `delta` refers to the difference between the mouse down coordinates and the mouse up coordinates, but we want the difference between the last mouse *drag* coordinates and the mouse up coordinates, in order to properly angle the end cap.

### DIFF
--- a/src/helper/blob-tools/broad-brush-helper.js
+++ b/src/helper/blob-tools/broad-brush-helper.js
@@ -79,7 +79,6 @@ class BroadBrushHelper {
                 this.endCaps.push(this.union(circ, this.union(rect, rect2)));
             }
         }
-        this.lastVec = event.delta;
         step.angle += 90;
 
         // Move the first point out away from the drag so that the end of the path is rounded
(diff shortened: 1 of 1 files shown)


fsih:
Awesome, I've been meaning to fix those little divets for forever

fsih (commented):
We decided this is already looking better than what we had before! But the simplify algorithm is occasionally giving sharp corners, and Max will look into it more

fsih:
The corners would require a lot of work to fix, so we'll look into it later

fsih (approved):
Looks much better, thanks!

PR #921, opened 2019-11-08
Skip processing of child paths in mergeEraser

adroitwhiz:
### Resolves

Resolves #621
Resolves #120
Resolves #119

### Proposed Changes

This PR modifies the shape Boolean processing loop in `Blobbiness.mergeEraser` to skip processing paths that are children of compound paths.

### Reason for Changes

Both child paths and their parents are included in the list of items that `mergeEraser` will attempt to merge with. If child paths are not specifically excluded, `mergeEraser` will attempt to operate on them twice: once with their parent path, and one with them on their own.

Fixing this greatly improves the reliability of eraser operations:

Before | After
-|-
mergeeraser-before | mergeeraser-after
mergeeraser-before-2 | mergeeraser-after-2

PR #1014, opened 2020-04-24
Remove compound path TODOs from blob.js

### LINKED ISSUE #119: Eraser inverts color/holes
eraserbug

This one is probably in paper.js. If you make a shape with a tail and a loop and subtract a shape that goes through both the tail and the loop, part of the path gets inverted.

adroitwhiz:
### Resolves

#119, I guess?

### Proposed Changes

Per [this comment](https://github.com/LLK/scratch-paint/pull/921#discussion_r414876979), this PR removes the `TODO` comments about handling compound paths in the `mergeEraser` code.

### Reason for Changes

Handling compound paths is now done properly, so the comments are no longer needed.

### DIFF
--- a/src/helper/blob-tools/blob.js
+++ b/src/helper/blob-tools/blob.js
@@ -272,7 +272,6 @@ class Blobbiness {
         }
 
         for (let i = items.length - 1; i >= 0; i--) {
-            // TODO handle compound paths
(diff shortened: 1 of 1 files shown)


fsih (approved):
Haha thanks

Issue #1078, opened 2020-05-25
rotationCenterX and rotationCenterY props should be optional


fsih:
### Expected Behavior
rotationCenterX and rotationCenterY props for the paint editor component are marked as optional, so the paint editor should work if you don't pass them in

### Actual Behavior
If you leave them out, you can't load bitmap images (nothing appears when you try)

### Steps to Reproduce
Don't pass in rotationCenterX or Y, and give the paint editor component a bitmap image. (Note that scratch-gui always passes in a rotation center, so the bug isn't visble)

### Operating System and Browser
all

BryceLTaylor:
This would be necessary to fix in order to add the ability to upload images to the playground, which can be found in this issue: https://github.com/LLK/scratch-paint/issues/1076

PR #1079, opened 2020-05-25
Add default rotation center for bitmap images

fsih:
### Resolves
https://github.com/LLK/scratch-paint/issues/1078

### Proposed Changes
If no rotation center is passed in, default to the center of the canvas

### Reason for Changes
it's better than nothing appearing

### Test Coverage
manual


ericrosenbaum (approved):
LGTM!

PR #1081, opened 2020-05-25
Set the width and height on the canvas used to convert imageData to dataURL

### LINKED ISSUE #1080: Playground code doesn't output bitmap images correctly
### Expected Behavior
The playground should be a working example of how to use scratch-paint

### Actual Behavior
Right now, the example code for outputing bitmap images is incorrect. It will always crop the output image to 300px x 150px

### Steps to Reproduce
In the playground, after drawing a bitmap image, log the result of this.state.image, and put the base64-encoded string into a base64 to image converter. Notice that the image is cropped.

all

fsih:
### Resolves
Resolves https://github.com/LLK/scratch-paint/issues/1080

### Proposed Changes
Set the width and height on the canvas used to convert imageData to dataURL

### Reason for Changes
By default, canvas size is 300px x 150px, which isn't usually the size you expect.

### Test Coverage
Tested manually

### DIFF
--- a/src/playground/playground.jsx
+++ b/src/playground/playground.jsx
@@ -43,6 +43,7 @@ class Playground extends React.Component {
             image: svgString, // svg string or data URI
             rtl: rtl
         };
+        this.reusableCanvas = document.createElement('canvas');
     }
     handleUpdateName (name) {
         this.setState({name});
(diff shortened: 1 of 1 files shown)


ericrosenbaum (approved):
LGTM! I tested on mac chrome, by getting the image data using the react chrome extension and pasting it into a base64 to image converter site

PR #1097, opened 2020-05-29
Make the room between canvas edges and scrollbars count as draggable

### LINKED ISSUE #1011: There's some room between the scrollbars and the canvas edges where you can interact with the canvas
### Expected Behavior

When you position your mouse around the scrollbar and drag, one of two things should happen:
- You drag the scrollbar
- Your mouse is off the edge of the canvas, and nothing happens

### Actual Behavior

There are a few pixels between the scrollbar and the edge of the canvas where you can accidentally do things that interact with the canvas:
Peek 2020-04-22 13-13

This is related to #611 somewhat-- because the scrollbars are so small, I often aim my mouse more towards the canvas edges than the center of the scrollbar, so as to avoid not getting close enough to the scrollbar and interacting with the canvas.

### Steps to Reproduce

1. Position the mouse in between the scrollbar and the edge of the canvas
2. Notice that dragging your mouse interacts with the canvas instead of the scrollbar

All

adroitwhiz:
### Resolves

Resolves #1011

### Proposed Changes

This PR changes the scrollbars to include the 4px margin between the scrollbar and canvas edge as part of the scrollbars' draggable area.

It does this by separating the scrollbar element into a "scrollbar hitbox" element and the actual drawn scrollbar.

### Reason for Changes

This avoids the frustration that occurs when you go to drag the scrollbar but position your mouse slightly too far down/to the right.

### Test Coverage

Tested manually

### DIFF
--- a/src/components/scrollable-canvas/scrollable-canvas.css
+++ b/src/components/scrollable-canvas/scrollable-canvas.css
@@ -2,31 +2,39 @@ $scrollbar-size: 8px;
 $scrollbar-padding: 4px;
 
 .vertical-scrollbar, .horizontal-scrollbar {
-    position: absolute;
     background: rgba(190, 190, 190, 0.8);
     border-radius: calc($scrollbar-size / 2);
-    cursor: pointer;
+    width: 100%;
+    height: 100%;
 }
 .vertical-scrollbar-wrapper {
     position: absolute;
-    width: $scrollbar-size;
-    right: $scrollbar-padding;
+    width: calc($scrollbar-size + $scrollbar-padding);
+    right: 0;
     top: $scrollbar-padding;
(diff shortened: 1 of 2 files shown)


fsih:
It looks like the scrollbars are filling the space now, rather than the empty space being clickable?
Screen Shot 2020-06-03 at 11 43 16

adroitwhiz:
@fsih What browser is that (Safari)? It seems to work on Chrome + Firefox

fsih:
I'm seeing it in all browsers on Mac. It looks like it's specific to the playground!

adroitwhiz:
It was a result of `box-sizing: border-box` being set for all elements in the GUI playground but not the paint editor playground. I explicitly applied it to the scrollbars here; it should now work in the paint editor playground

fsih (approved):
LG, thanks!

PR #1100, opened 2020-05-29
Delete item.data.origPos instead of setting it to null

### LINKED ISSUE #1099: Dragging shapes around adds pointless extra attributes to exported SVG
### Expected Behavior

Exported SVGs should not contain any unnecessary attributes/data

### Actual Behavior

In vector mode, if you drag an item around in the paint editor, `data-paper-data="{&quot;origPos&quot;:null}"` will be added to its corresponding exported SVG element.

### Steps to Reproduce

1. Ungroup the default cat costume
2. Drag some of its shapes around
3. Export the costume
4. Open the exported SVG in a text editor and search for `data-paper-data`
5. Notice that it appears on every element you dragged around in the paint editor

All

adroitwhiz:
### Resolves

Resolves #1099

### Proposed Changes

This PR changes the move tool to delete `item.data.origPos` on its selected items when finished dragging, as opposed to setting it to `null` as it previously did.

### Reason for Changes

Dragging a vector shape causes the `MoveTool` to set the `origPos` property on that item's `data`. Once the dragging is done, `origPos` was previously then set to `null`.

However, since it was still present in the item's `data`, paper.js was serializing it into the saved SVG.

Now, the `MoveTool` deletes the `origPos` property entirely. This means that paper.js won't serialize it, decreasing the size of saved SVGs.

### Test Coverage

Tested manually

### DIFF
--- a/src/helper/selection-tools/move-tool.js
+++ b/src/helper/selection-tools/move-tool.js
@@ -206,10 +206,10 @@ class MoveTool {
         let moved = false;
         // resetting the items origin point for the next usage
         for (const item of this.selectedItems) {
-            if (item.data && item.data.origPos && !item.position.equals(item.data.origPos)) {
-                moved = true;
+            if (item.data.origPos) {
+                if (!item.position.equals(item.data.origPos)) moved = true;
+                delete item.data.origPos;
             }
-            item.data.origPos = null;
         }
         this.selectedItems = null;
         this.selectionCenter = null;


src/helper/selection-tools/move-tool.js
```diff
@@ -209,7 +209,7 @@ class MoveTool {
             if (item.data && item.data.origPos && !item.position.equals(item.data.origPos)) {
                 moved = true;
             }
-            item.data.origPos = null;
+            delete item.data.origPos;
```
fsih: Should this move into if (item.data && item.data.origPos) ?

fsih: Alternatively maybe the above if doesn't need to check for those two things?

adroitwhiz: I removed the check for `item.data` since we were unconditionally accessing `item.data.origPos` anyway, and moved the `delete` inside the conditional.

fsih (approved):
LG! Thanks!

PR #1114, opened 2020-06-07
Undo/Redo should not be reversed in RTL languages

### LINKED ISSUE #987: Undo/Redo should not be reversed in RTL languages
### Expected Behavior

In general things associated with time still go left-to-right in RTL languages. Since Undo/Redo are "time" related, they should stay in their usual orientation (Undo on the left, arrow pointing left/Redo on the right, arrow pointing right).

### Actual Behavior

Position and arrow direction is reversed in RTL langauges.

### Steps to Reproduce

Switch to Hebrew or Persian (Farsi).

Note, this was reported by a Persian speaker. We should verify that the direction makes sense for other RTL languages (e.g. Hebrew).

Jacco:
### Resolves

Fixes #987 

According to my research imagery needs to stay in position but functionality and labeling needs to be reversed in RTL.

[link to article](https://medium.com/@giladalmosnino/ux-best-practices-for-bi-directional-languages-9bd7b96dc8c2)

### Reason for Changes

See issue

### Test Coverage

I did not add a test

### DIFF
--- a/src/components/fixed-tools/fixed-tools.jsx
+++ b/src/components/fixed-tools/fixed-tools.jsx
@@ -125,11 +125,11 @@ const FixedToolsComponent = props => {
                                 }
                             )
                         }
-                        disabled={undoDisabled}
-                        onClick={props.onUndo}
+                        disabled={props.rtl ? redoDisabled : undoDisabled}
+                        onClick={props.rtl ? props.onRedo : props.onUndo}
                     >
                         <img
-                            alt={props.intl.formatMessage(messages.undo)}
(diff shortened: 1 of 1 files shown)


chrisgarrity:
@Jacco on further research, we decided that the original issue was probably incorrect. Other applications, e.g. Google Docs, do mirror the undo/redo buttons so we do not need to change what is currently implemented. Thanks for taking a look at this in any case!

Jacco:
@chrisgarrity I left a comment in the issue #987

PR #1119, opened 2020-06-12
Fix color button turning white when it shouldn't

### LINKED ISSUE #1118: Selecting a group with "mixed" fill color whose items have gradients breaks the color swatch preview
### Expected Behavior

The fill selector should always show a preview of the current color.

### Actual Behavior

If you select a group which contains multiple shapes, some of which have gradient fills, and which has a "mixed" fill color (causes the fill selector to look like image) it will cause the fill selector to always appear white when you select items with solid colors, until you switch to a shape tool and switch back.

Peek 2020-06-12 02-41

### Steps to Reproduce

1. Enter the paint editor in vector mode
2. Create a new costume
3. Draw two circles
4. Switch to the select tool and set the gradient type of the first circle to something other than "solid"
5. Change the color of the second circle, then set its gradient type as well
6. Select both circles and group them
7. Draw another circle
8. Select the group of two circles
9. Select the third circle
10. Notice that the fill color selector is white

All

adroitwhiz:
### Resolves

Resolves #1118

### Proposed Changes

This PR changes the "color is mixed" check in `colorToBackground` to match the one in the `ColorButton` component itself.

Previously, `colorToBackground` considered a color state to be "mixed" if `color` was `MIXED` or `color2` was `MIXED`. Now, it considers a color state "mixed" if `color` is `MIXED` or `color2` is `MIXED` *and* the gradient type is non-solid.

### Reason for Changes

If you select a group with "mixed" gradient types, then color and color2 will both be MIXED. If you then select a shape with a solid color, it won't reset color2, which will remain MIXED. We only want the background to appear white if color2 is MIXED *and* the gradient type actually uses color2. This is already done in the component itself, which only shows the "mixed" icon if the gradient type is non-solid, but colorToBackground left out this gradient type check.

### Test Coverage
Tested manually

### DIFF
--- a/src/components/color-button/color-button.jsx
+++ b/src/components/color-button/color-button.jsx
@@ -11,7 +11,7 @@ import GradientTypes from '../../lib/gradient-types';
 import log from '../../log/log';
 
 const colorToBackground = (color, color2, gradientType) => {
-    if (color === MIXED || color2 === MIXED) return 'white';
+    if (color === MIXED || (gradientType !== GradientTypes.SOLID && color2 === MIXED)) return 'white';
     if (color === null) color = 'white';
     if (color2 === null) color2 = 'white';
     switch (gradientType) {


fsih (approved):
Thanks, LGTM!

PR #1120, opened 2020-06-13
prevent scaling factor too low (per direction)

### LINKED ISSUE #379: A line will disappear in the paint editor if one of its corner points is dragged to flip the line.
### Expected behavior
Lines can be flipped over using one of the corner selection vertices. 

### Actual behavior
The line disappears.

### To reproduce:
Mouse over Sprite Library and click on paint.
Click the line tool and draw a horizontal line with a slight slope.
Select the selection tool
Select the line
Click and hold one of the overlapping corner points and drag towards the line or sometimes away from the line.
Notice that it will disappear as you flip the line or increase the length.

Jacco:
### Resolves

#379 

### Proposed Changes

Keep sx factors on lastSx if it comes below 0.001
Keep sy factors on lastSy if it comes below 0.001

### Reason for Changes

If scaling factors become too low they become unrecoverable because of double precision and zero divides.

### Test Coverage

I did not create tests

### DIFF
--- a/src/helper/selection-tools/scale-tool.js
+++ b/src/helper/selection-tools/scale-tool.js
@@ -3,6 +3,8 @@ import {getItems} from '../selection';
 import {getActionBounds} from '../view';
 import {BitmapModes} from '../../lib/modes';
 
+const MIN_SCALE_FACTOR = 0.0001;
+
 /**
  * Tool to handle scaling items by pulling on the handles around the edges of the bounding
  * box when in the bounding box tool.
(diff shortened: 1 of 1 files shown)


src/helper/selection-tools/scale-tool.js
```diff
@@ -121,6 +121,12 @@ class ScaleTool {
             sx *= signx;
             sy *= signy;
         }
+        if (Math.abs(sx) < 0.0001) {
```
fsih: Thanks! This seems like a good way to fix the issue!
Rather than retaining lastSx, should sx snap to +-.0001?

fsih (changes requested):

src/helper/selection-tools/scale-tool.js
Jacco: @fsih I think that would also be possible, the 0.0001 was chosen arbitrary. The difference between the two would normally not be noticeable I think. I can change it if you want. Let me know.

I also wanted to add restricting to whole x/y increment (so only 100%, 200% etc for width cq height) using the ctrl key as a modifier. I know that handy modifier from Corel Draw! in the old days :-) Would you consider a feature like that? Shall I create an issue for that?

fsih: Hmm, I agree that it's very unlikely to be noticeable, but let's snap to 0.0001 so that behavior is always consistent. Can you also define a constant for .0001, maybe MIN_SCALE_FACTOR?

That feature from Corel Draw sounds handy! Sure, if you create an issue, I can ask our design team what they think.

fsih (approved):
Thank you, this will help a lot!

PR #1128, opened 2020-06-16
Make generated secondary color always visibly differ from primary color

### LINKED ISSUE #1127: Automatically-chosen secondary colors don't always differ from the primary color
### Expected Behavior

If you choose a gradient type other than "solid" in the color picker, the secondary color should visibly differ from the primary color, in order to make the gradient effect more visible

### Actual Behavior

The secondary color is calculated by altering the hue of the primary color. This means that if the primary color has 0 saturation, then this hue shift will have no visible impact and the secondary color will be the same as the primary color.

### Steps to Reproduce

1. Enter the paint editor
2. Open the fill selector
3. Drag the "Saturation" slider to 0
4. (optional) Drag the "Brightness" slider to some arbitrary value
5. Click on one of the gradient buttons to change the gradient type
6. Notice that both gradient colors are the same

Peek 2020-06-16 12-09

All

adroitwhiz:
### Resolves

Resolves #1127

### Proposed Changes

This PR changes the way corresponding secondary colors for gradients are chosen:

- If the primary color is bright and saturated, it will return that color, hue-shifted (like it currently does)
- If the primary color is dark and saturated, it will return that color, hue-shifted and brightened
- If the primary color is not very saturated, it will return that color, brightened or darkened (whichever contrasts more)

Note that if the primary color is not saturated, then the secondary color will not be hue-shifted. I did this because a hue shift would be hard to see anyway, and might create confusion, but it may also be confusing to only hue-shift *sometimes*? I'm not really sure.

### Reason for Changes

This should make gradients a lot more obvious when you enable them under certain circumstances (e.g. your primary color is dark or desaturated).

In particular, since the default outline color is black, enabling gradients on outlines would otherwise be very non-obvious/confusing.

### Test Coverage

Tested manually

### DIFF
--- a/src/containers/bit-fill-mode.jsx
+++ b/src/containers/bit-fill-mode.jsx
@@ -13,7 +13,7 @@ import {clearSelectedItems} from '../reducers/selected-items';
 import {changeGradientType} from '../reducers/fill-mode-gradient-type';
 import {clearSelection} from '../helper/selection';
 import FillTool from '../helper/bit-tools/fill-tool';
-import {getRotatedColor, MIXED} from '../helper/style-path';
+import {generateSecondaryColor, MIXED} from '../helper/style-path';
 
 class BitFillMode extends React.Component {
     constructor (props) {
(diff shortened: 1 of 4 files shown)


fsih (dismissed):
LG, thanks!

fsih (approved):

PR #1155, opened 2020-06-28
#1150 Hor/ver line scaling

### LINKED ISSUE #1150: Scaling doesn't work as expected on perfectly horizontal/vertical lines
horizontal line

Jacco:
### Resolves

Fixes #1150 

### Proposed Changes

Do not show scaling anchors when the anchor would scale a zero sized side.

This means for
- zero height () bounds only show left-middle and right-middle handles
- zero width () bounds only show top-middel and bottom-middle handles

Also I corrected the use of boundsPath.bounds ScaleTool. boundsPath was including the selectionAnchor as a child, so the size of the anchor (24) was also used in the calculations. I fixed this by adding selectionRect separate of selectionAnchor to the boundsPath.

(I used 1e-8 as zero bound because I saw one case of a vertical line where width of bounds was 1e-14)

### Reason for Changes

To fix issue

### Test Coverage

Manually tested


ilikecereal1legacy:
any developments?

PR #1156, opened 2020-06-28
fix 1132

### LINKED ISSUE #1132: Another scale modifier key for "whole multiples": CTRL
### Expected Behavior

Support another constrained resize mode keeping the width / height to whole multiple of the original. I know this behavior to be very handy from the days I worked with Corel DRAW! The modifier was under the CTRL key. In Corel DRAW the aspect constraint was not active by default but rather activated with shift. This made it easier to remember what the constrainer keys do.

CTRL (constrains aspect ratio AND constrains to multiples)

ctrl

The modifiers combine nicely:

CTRL + ALT (constrains aspect ratio AND constrains to multiples AND keeps center in place):

withalt

CTRL + SHIFT (constrains to multiples):

withshift

CTRL + ALT + SHIFT (constrains to multiples AND keeps center in place)

withshiftalt

### Actual Behavior

This behavior is currently not supported.

### Steps to Reproduce

This behavior is currently not supported.

It concerns a new feature.

Jacco:
### Resolves

Fixes #1132 

### Proposed Changes

Added rounding of sx/sy when ctrl modifier is used.

I also moved splitting of sizing factors into sign and absolute value since the abs-ing would otherwise be done separately 3 times.

### Reason for Changes

Its handy :-)

### Test Coverage

Manually tested

### DIFF
--- a/src/helper/selection-tools/scale-tool.js
+++ b/src/helper/selection-tools/scale-tool.js
@@ -118,13 +118,19 @@ class ScaleTool {
 
         const signx = sx > 0 ? 1 : -1;
         const signy = sy > 0 ? 1 : -1;
+        sx = Math.abs(sx);
+        sy = Math.abs(sy);
         if (this.isCorner && !event.modifiers.shift) {
-            sx = sy = Math.max(Math.abs(sx), Math.abs(sy));
-            sx *= signx;
-            sy *= signy;
+            sx = sy = Math.max(sx, sy);
         }
-        sx = signx * Math.max(Math.abs(sx), MIN_SCALE_FACTOR);
-        sy = signy * Math.max(Math.abs(sy), MIN_SCALE_FACTOR);
+        if (event.modifiers.control) {
+            sx = Math.max(Math.round(sx), 1);
+            sy = Math.max(Math.round(sy), 1);
+        }
+        sx = Math.max(sx, MIN_SCALE_FACTOR);
+        sy = Math.max(sy, MIN_SCALE_FACTOR);
+        sx *= signx;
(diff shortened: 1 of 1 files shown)


adroitwhiz:
It may be useful to scale by integer increments (2x, 3x, 4x, 5x, etc) when scaling up, and powers of negative two (0.5x, 0.25x, 0.125x, etc) when scaling down. For instance, bitmap costumes are imported at double resolution, and users may desire a way to scale them down by exactly half.

fsih:
@Jacco I think @adroitwhiz's proposal for making it snap down to 1/2^x increments sounds good. What do you think of making that change?

We can pick a minimum power of 1/2 that's greater than your minimum scale value, or change the minimum scale value to a power of 2, so that there aren't needlessly many "stops" near 0

Jacco:
@fsih @adroitwhiz Would love to make this addition. But I have one request before that.

It would be great to have a place in the editor to show some details about the current tool action.

Reason is that for the smaller powers of 2 (say 1/32 1/64 1/128) it will be very hard to see which one is the current one.

For example for the scaling tool it could display:
tool used: scale (maybe icon)
modifiers: ("centered" and/or "free aspect" and/or "constrained") (maybe icons?)
details: original width x height -> new width x height
constrained snap details: for example (snapped to 2 x 1/2)

fsih:
@Jacco our designers think that displaying the scale metrics would be a great idea! Unfortunately, changes to the UI need to go through design iteration and playtesting with a group of kids, and we don't have the design bandwidth for that right now. Could you file that as a new feature request issue?

PR #1166, opened 2020-07-09
Gradient outlines part 1

adroitwhiz:
### Resolves

A step towards https://github.com/LLK/scratch-paint/issues/540

### Proposed Changes

This PR implements the "refactoring" half of the "gradient outlines" changes--this should contain no functional changes but cleans up the codebase in preparation for adding gradient outline support.

### Reason for Changes

See above

### Test Coverage

Tests have been updated accordingly


fsih (changes requested):
Thanks for splitting all this out, it helps a lot!

I ran into an issue where if you select a shape with a hole, and set the fill to a gradient, then the fill becomes transparent:
Screen Shot 2020-07-13 at 15 55 33

fsih:
Here's the same SVG exported from scratch.mit.edu vs exported from the gradient-outlines branch: https://www.diffchecker.com/xGHcgGmX

src/helper/style-path.js
```diff
@@ -339,6 +350,33 @@ const applyStrokeWidthToSelection = function (value, textEditTargetId) {
...
+            null :
+            gradient.stops[1].color.toCSS();
+    } else {
+        if (gradient.stops.length < 2) log.warn(`Gradient has ${gradient.stops.length} stop(s)`);
+
+        colorState.primary = MIXED;
+        colorState.secondary = MIXED;
+    }
+
+    return colorState;
```
fsih, line 358: Can we document the params of the return value, and note that gradientType may be null?

fsih (dismissed):
LGTM!

adroitwhiz:
I've changed `applyGradientTypeToSelection` to call `createGradientObject` when creating gradients-- this is important in part 2 when we change `createGradientObject` to enforce a minimum distance between the gradient start and end points (this fixes perfectly vertical gradients on perfectly horizontal lines, or vice versa, from disappearing).

fsih (approved):
LG!

PR #1197, opened 2020-07-29
Don't clear gradient on entering vector line mode

adroitwhiz:
### Resolves

N/A (should I bother making one?)

### Proposed Changes

Stop clearing the gradient when entering vector line mode

### Reason for Changes

We now support gradients in line mode

### DIFF
--- a/src/containers/line-mode.jsx
+++ b/src/containers/line-mode.jsx
@@ -9,7 +9,7 @@ import {clearSelection} from '../helper/selection';
 import {endPointHit, touching} from '../helper/snapping';
 import {drawHitPoint, removeHitPoint} from '../helper/guides';
 import {styleShape} from '../helper/style-path';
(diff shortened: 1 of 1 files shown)


src/containers/line-mode.jsx
```diff
@@ -294,9 +291,6 @@ const mapStateToProps = state => ({
     isLineModeActive: state.scratchPaint.mode === Modes.LINE
 });
 const mapDispatchToProps = dispatch => ({
-    clearGradient: () => {
-        dispatch(clearStrokeGradient());
```
fsih, line 298: Looks like this function might not be used anywhere anymore, and we could remove it?

fsih: ok we'll just leave it for now since it's connected to the generic "color state" reducer generator

fsih (approved):
LG!

PR #1203, opened 2020-07-31
Make slider handles stick to the cursor

### LINKED ISSUE #1202: Color sliders "jump" when dragged
### Expected Behavior

Clicking and dragging a color slider should be a smooth experience

### Actual Behavior

When you begin to drag a color slider, the handle will "jump" to a certain position, causing a rapid change in the slider's value:
Peek 2020-07-31 12-26

### Steps to Reproduce

1. Open the color picker
2. Mouse down on a color slider handle
3. Move the mouse slightly and notice that the slider has suddenly changed position too much

All

adroitwhiz:
### Resolves

Resolves #1202

### Proposed Changes

This PR adjusts the color slider position calculation to take into account where the user first clicked on the handle.

### Reason for Changes

This prevents sliders from "jumping" when the user starts dragging them:

Before | After
--- | ---
Peek 2020-07-31 12-26 | Peek 2020-07-31 12-32

### Test Coverage

Tested manually.

### DIFF
--- a/src/components/color-picker/color-picker.jsx
+++ b/src/components/color-picker/color-picker.jsx
@@ -5,7 +5,7 @@ import {defineMessages, FormattedMessage, injectIntl, intlShape} from 'react-int
 import classNames from 'classnames';
 import parseColor from 'parse-color';
 
-import Slider from '../forms/slider.jsx';
+import Slider, {CONTAINER_WIDTH, HANDLE_WIDTH} from '../forms/slider.jsx';
 import LabeledIconButton from '../labeled-icon-button/labeled-icon-button.jsx';
 import styles from './color-picker.css';
 import GradientTypes from '../../lib/gradient-types';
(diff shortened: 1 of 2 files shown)


fsih:
I'm seeing an issue where the handle jumps a bit when I mouse up
jump

adroitwhiz:
The slider jumping on mouse up has been fixed.

There's another issue that I introduced and have fixed as well, and it affects the way sliders are displayed. Essentially, because we're treating the center of the slider handle as the value, the slider's rounded end caps are outside its effective "range":
image

When the slider is all the way to the left, for instance, the center of the slider handle is still 13 pixels to the right of where the slider starts.

Because of this, I changed the way the slider gradients are displayed so that the rounded end caps are solid, and the gradient only transitions in the region shown above.

fsih (approved):
Looks great! Thanks!

PR #1211, opened 2020-08-09
Gradient outlines fixes

fsih:
### Resolves
Fixes a couple of gradient outlines blockers
1) If the first color of a gradient outline is transparent, the outline width appears to be 0
https://github.com/LLK/scratch-paint/issues/1214

2) The second color of the gradient outline can appear to be mixed
https://github.com/LLK/scratch-paint/issues/1215

### Proposed Changes
1. Change the color to black in line mode only if the primary color is mixed or both the primary and secondary color are undefined
Clear the stroke gradient if the second stroke color is mixed
(Before, it would let the secondary color be mixed in line mode)
2nd mixed

2. When applying a non-solid gradient type to a transparent, solid outline color, the second color is supposed to be a non-transparent color, and the outline width is supposed to change to 1 (since it's now not totally transparent) (before, it would stay 0)
when yu add a gradient

3. Both the primary and secondary colors should be transparent before we force the stroke width to 0. (Before, if only the first color was transparent, the stroke width would become 0)
transparent is 0

4. Allow gradients from transparent to transparent to still show up as gradients in the fill dropdown, otherwise when we switch from that to a gradient, it doesn't force a second color/outline width to be added.
Before:
transparent to transparent before

After:
transparent to transparent

### Reason for Changes

### Test Coverage

### DIFF
--- a/src/containers/line-mode.jsx
+++ b/src/containers/line-mode.jsx
@@ -280,6 +280,7 @@ class LineMode extends React.Component {
 
 LineMode.propTypes = {
     clearSelectedItems: PropTypes.func.isRequired,
+    clearStrokeGradient: PropTypes.func.isRequired,
     colorState: PropTypes.shape({
         fillColor: ColorStyleProptype,
         strokeColor: ColorStyleProptype,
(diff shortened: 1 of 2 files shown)


fsih:
@adroitwhiz if you want to double check my work, that would be really reassuring!

ericrosenbaum (approved):
LGTM!

PR #1280, opened 2020-10-14
Remove dependency on svg renderer

fsih:
### Proposed Changes
Remove the dependency on scratch-svg-renderer from paint.

https://github.com/LLK/scratch-gui/pull/6262 is the gui side of this change.

### Reason for Changes
svg-renderer is difficult to test and hotfix because it's so deep in the dependency chain. This is for tech ergonomics.

### Test Coverage
Tested the playground and linked to scratch-gui


cwillisf (approved):
:+1:

PR #1293, opened 2020-10-22
Add scratch render fonts as a peer dependency

fsih:
### Resolves
This dep should have been in scratch paint's peer dependencies, because although paint works fine without it, the fonts would not render when using the text tool without this peer dependency

Currently scratch-render-fonts comes in through scratch-svg-renderer, but this will change with https://github.com/LLK/scratch-paint/pull/1280


fsih:
Moving to https://github.com/LLK/scratch-paint/pull/1280

PR #210, opened 2017-12-05
Show the bounding box after switching to select tool and transforming

fsih:
Fix https://github.com/LLK/scratch-paint/issues/200

Some changes in https://github.com/LLK/scratch-paint/commit/d918fc0bdaf3e1e06a328f01cafe6bf1a6e4367a were too aggressive

PR #1346, opened 2021-01-07
Use fixupSVGString from scratch-svg-renderer

### LINKED ISSUE #1269: SVG fixes should be moved to svg-renderer's fix-up-svg
### Expected Behavior
SVG fixes are performed in svg-renderer, where they are fixed for assets across scratch repos

### Actual Behavior
Some fixes are performed only in Paint: https://github.com/LLK/scratch-paint/blob/7dd437415c8dcc921d68627c26af1c6e423f2d70/src/containers/paper-canvas.jsx#L197

Moving the fixes to svg-renderer can help fix security issues

adroitwhiz:
### Resolves

Resolves #1269

### Proposed Changes

This PR changes the `importSVG` function to use the `fixupSvgString` function from `scratch-svg-renderer` instead of applying its own fixes to the imported SVG string.

### Reason for Changes

See the rationale in #210. This should result in all renderable SVGs also being editable in the paint editor--no more SVGs that render but crash the paint editor. Now, either an SVG is parseable or it'll be replaced with the grey question mark asset everywhere.

### Test Coverage

Tested manually

### DIFF
--- a/src/containers/paint-editor.jsx
+++ b/src/containers/paint-editor.jsx
@@ -310,6 +310,7 @@ class PaintEditor extends React.Component {
                 canUndo={this.props.shouldShowUndo}
                 canvas={this.state.canvas}
                 colorInfo={this.state.colorInfo}
+                fixupSvgStringFn={this.props.fixupSvgStringFn}
                 format={this.props.format}
                 image={this.props.image}
                 imageFormat={this.props.imageFormat}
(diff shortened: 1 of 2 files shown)


benjiwheeler:
Update: we're doing a similar thing now, with a new sanitization function that is exported from scratch-svg-renderer and which is run on image upload. There are still some differences between rendering in the stage/tiles and in the paint editor, however -- I'm not sure where the problem is with those.

PR #1386, opened 2021-02-09
Update to most recent eslint-config-scratch and associated fixes

paulkaplan:
This change updates the version of `eslint-config-scratch` to the most recent version. That version introduces a few new es6 specific rules, like not allowing duplicate imports, so I had to fix all those first. 

I imagine these duplicate imports accumulated over time, but just to be clear its ok to import both default _and_ named imports at the same time. Most of these changes are of the form

### no-duplicate-imports
```diff
-import DefaultThing from 'somewhere';
-import {NamedThing} from 'somewhere';
+import DefaultThing, {NamedThing} from 'somewhere';
```

Although `eslint-config-scratch@6.0.0` only requires eslint 5.3.0, i upgraded to 5.7.0 just because I know that changes that require that version are in progress https://github.com/LLK/eslint-config-scratch/pull/76 But these fixes are required and separate from that update, so figured I might as well update it now.


fsih (approved):
Thanks Paul! Yeah, I didn't know about that back then

PR #1388, opened 2021-02-10
Mark deprecated lifecycle methods

paulkaplan:
NB this depends on https://github.com/LLK/scratch-paint/pull/13862, the only new commit here is 5850c40

This is exactly the same process as with GUI https://github.com/LLK/scratch-gui/pull/6723

As preparation for updating to react 17, mark deprecated lifecycle methods as UNSAFE_ to reduce the console warnings. Once we update to react 17, we can migrate to newer lifecycle methods or otherwise remove the UNSAFE_ methods.

Include the correct camelcase rule in the eslintrc, since it is only temporary.


apple502j:
Oh gosh this all depends on deprecated stuff.

I think from the code it seems like it all causes side effects, though. componentDidUpdate could be worth a try, but that could cause some inconsistencies because of props and paper state mismatch. Hmmm.

paulkaplan:
As mentioned in https://github.com/LLK/scratch-gui/pull/6723, going to hold on this because these methods aren't availlable in 16.2

cwillisf:
This might be good reference material when we try again to upgrade React

PR #1464, opened 2021-03-17
Fix fill tool for touch: update hover on mouse up

### LINKED ISSUE #898: On touch devices, paint bucket only fills if you drag, not if you tap
Observed on Windows Edge touch (by @benjiwheeler), iPad Safari (by @benjiwheeler), and Android Chrome (by @chrisgarrity)

Steps:
create default project
edit scratch cat in paint editor, in vector mode
select paint bucket and some fill color
tap on cat in various places; most taps won’t result in a fill; all taps where I drag a bit first do fill.

cwillisf:
### Resolves

Resolves #898 

### Proposed Changes

In the fill tool, update hover info by calling `handleMouseMove` at the beginning of `handleMouseUp`.

### Reason for Changes

The `handleMouseUp` function applies the fill tool to the object in `this.fillItem`, which is set by `handleMouseMove`. This assumes that `handleMouseMove` will be called at least once before `handleMouseUp` is called. That's generally a safe assumption when using a mouse, but touch down/up events can happen anywhere at any time, often without a preceding move event. In that case, we still need to figure out what the touch is "over" in order to fill the right item. Pretending that the mouse-up is a mouse-move updates that information.

### Test Coverage

Tested locally in the scratch-paint playground.

### DIFF
--- a/src/helper/tools/fill-tool.js
+++ b/src/helper/tools/fill-tool.js
@@ -21,6 +21,7 @@ class FillTool extends paper.Tool {
 
         // We have to set these functions instead of just declaring them because
         // paper.js tools hook up the listeners in the setter functions.
+        this.onMouseDown = this.handleMouseDown;
         this.onMouseMove = this.handleMouseMove;
         this.onMouseUp = this.handleMouseUp;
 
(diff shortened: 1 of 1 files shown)


fsih (approved):
Thanks for finally fixing this!!!

PR #1638, opened 2021-06-08
Enable semantic release

fsih:
Semantic release will help prevent broken dependency problems


adroitwhiz:
By default I think commits marked as 'refactor' won't trigger a release; not sure if that's intentional

apple502j:
Also, why are you reverting the version? That's not how semver works? (And probably messes up NPM)

cwillisf:
This PR is being replaced by #2052, but to answer your question about the version number: `semantic-release` calculates the version number on its own and ignores `package.json`.

This version of the PR would have stuck the version number at `0.0.0-development` to emphasize that the version number in `package.json` isn't the "real" version number.

The new PR configures `semantic-release` to update `package.json` with the calculated version, which I think is even better.

PR #1882, opened 2022-03-08
move browserslist to package.json

BryceLTaylor:
### Resolves
We're getting some errors in the build process

### Proposed Changes

Move the browser list from being an argument for autoprefixer int he webpack config to the package.json

### Reason for Changes
remove build errors

### Test Coverage

Tested in Chrome and Safari

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,21571 @@
+{
+  "name": "scratch-paint",
+  "version": "0.2.0",
+  "lockfileVersion": 1,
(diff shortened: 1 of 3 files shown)


cwillisf (approved):
This change also adds `package-lock.json`, which wasn't checked in before. Is that intentional?
If so, this all looks good! I'll approve on the assumption that it's intentional, but if not it might be better to remove it from this PR and add it separately (or at least in a separate commit)

PR #1930, opened 2022-03-28
Setup initial circleCI config with build-test job

BryceLTaylor:
### Resolves

### Proposed Changes

Sets up the initial config for circleci
Initially circle will only build and not deploy
Circle will run in conjunction with Travis for the time being
Ultimately Circle will work better with Github for deploy purposes.

### Reason for Changes

This will begin the process of transitioning from TravisCI to CircleCI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,49 @@
+version: 2.1
+aliases:
+  - &defaults
+      docker:
+        - image: cimg/node:12-browsers
+      working_directory: ~/repo
+      environment:
+        NODE_ENV: production
+        RELEASE_TIMESTAMP: "$(date +'%Y%m%d%H%M%S')"
+  - &setup
+    name: "setup"
+    command: |
+      npm --production=false ci
+      mkdir ./test/results
+  - &lint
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,44 @@
...
+  build-test:
+    <<: *defaults
+    steps:
+      - checkout
+      - run:
+          <<: *setup
+      - run:
+          <<: *lint
+      - run:
+          <<: *build
```
cwillisf, line 38: Seems like this should also run `npm run unit`

cwillisf (changes requested):
Seems like `build-test` should run the tests, but other than that -- or if skipping that is intentional for some reason... looks good to me!

.circleci/config.yml
BryceLTaylor: Yep.  I just updated it.  Good catch.  I think I missed it because Travis calls Test automatically so I didn't noticed that Test calls Unit.

cwillisf (approved):
Thanks!

PR #1959, opened 2022-04-19
circleci ignores gh-pages branches

BryceLTaylor:
### Resolves

pushing to gh-pages kicks off another build-test workflow, which happens every time we deploy

### Proposed Changes

adds the gh-pages branch to the list of branches to be ignored in the circleci config

### Reason for Changes

Will make it so we don't kick off an extra build-test workflow after we deploy


cwillisf (approved):
Looks good!

PR #1960, opened 2022-04-20
Hotfix/circleci deploy npm

BryceLTaylor:
### Proposed Changes

circleci can now deploy to npm and push the release tag to github as part of the deploy

### Reason for Changes

To move completely from travis to circleci we need to deploy to npm.  Part of that deploy is to tag the correct commit with the release version in github

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,13 +29,11 @@ aliases:
     command: |
       RELEASE_TIMESTAMP="$(date +'%Y%m%d%H%M%S')"
       VPKG=$($(npm bin)/json -f package.json version)
-      export RELEASE_VERSION=${VPKG}-prerelease.${RELEASE_TIMESTAMP}
-      echo $RELEASE_VERSION
+      echo export RELEASE_VERSION=${VPKG}-prerelease.${RELEASE_TIMESTAMP} >> $BASH_ENV
       echo export NPM_TAG=latest >> $BASH_ENV
       if [[ "$CIRCLE_BRANCH" == hotfix/* ]]; then # double brackets are important for matching the wildcard
         echo export NPM_TAG=hotfix >> $BASH_ENV
       fi
-      npm version --no-git-tag-version $RELEASE_VERSION
   - &deploy-gh-pages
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -29,18 +29,31 @@ aliases:
...
+      echo "release version: $RELEASE_VERSION"
+      npm version --no-git-tag-version $RELEASE_VERSION
+      npm set //registry.npmjs.org/:_authToken=$NPM_TOKEN
+      npm publish --tag $NPM_TAG
+  - &tag-commit
+    name: "tag commit in github"
+    command: |
+      echo $RELEASE_VERSION
+      git tag $RELEASE_VERSION
+      git push $CIRCLE_REPOSITORY_URL $RELEASE_VERSION
```
cwillisf, line 56: It's a minor point, but I feel like we don't need to fight `npm` with `--no-git-tag-version` here. I think a roughly-equivalent set of steps would be to flip the order that `tag-commit` and `deploy-npm` steps run in, and adjust them to look something like this:

```suggestion
      echo "release version: $RELEASE_VERSION"
      npm set //registry.npmjs.org/:_authToken=$NPM_TOKEN
      npm publish --tag $NPM_TAG
  - &tag-commit
    name: "tag commit in github"
    command: |
      echo $RELEASE_VERSION
      npm version $RELEASE_VERSION
      git push $CIRCLE_REPOSITORY_URL $RELEASE_VERSION
```

(I didn't flip the order here because I wanted the diff to be easier to read.)

In that case I think it would also make sense to rename `tag-commit` to something like `bump-version`.

cwillisf (approved):
This looks good! I could see a slightly different way to do it, which I shared in a comment. Up to you if you want to try that out.

.circleci/config.yml
BryceLTaylor: The way it's set up currently matches how we were doing it in Travis, and also more resembles how it works in GUI.  It might be worth coming back and changing our approach later, but for now I think we should leave it as is.

PR #1964, opened 2022-04-22
Circleci setup cron to push translations

BryceLTaylor:
### Proposed Changes

Adds a workflow to circleci that pushes translations to Transifex.

### Reason for Changes

This is the last piece that Travis handled for scratch-paint


cwillisf (approved):
I've never used triggers before on CirleCI, but this LGTM!

PR #1969, opened 2022-04-27
Remove Travis.yml

BryceLTaylor:
No longer needed because we have CircleCI set up

### Proposed Changes

Removes the Travis.yml so we won't use it anymore.

### Reason for Changes

All functionality from Travis has been moved into Circleci, so this is no longer needed

### Test Coverage

CircleCI runs unit tests already.


cwillisf (approved):
🚀 🟢

PR #1972, opened 2022-04-29
Use published gh-pages NPM package

rschamp:
The fork we were using is no longer necessary, as the “publish to subdirectory” `-e` option has been integrated into the official package.

scratch-paint version of https://github.com/LLK/scratch-gui/pull/8258

### Resolves

Doesn't resolve any issue, but removes dependency between this repo and the @rschamp account.

### Proposed Changes

Updates the gh-pages dependency from a github dependency to the up to date NPM dependency for that package.

### Reason for Changes

We weren't getting any updates for gh-pages while it was sitting on my fork. And now the reason for using the fork is no longer true: The published package has the -e option.

### Test Coverage

I pushed this branch to LLK, so we can see if it pushes to gh-pages after building

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2183,12 +2183,20 @@
       "dev": true
     },
     "async": {
-      "version": "2.1.2",
-      "resolved": "https://registry.npmjs.org/async/-/async-2.1.2.tgz",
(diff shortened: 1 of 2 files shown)


scratch-deployer (approved):
Looks good!

PR #2052, opened 2022-08-25
Semantic release

### LINKED ISSUE #1588: npm-published packages should follow semver
### Expected Behavior

Scratch projects published to npm should use semver version numbers >= 1.0 so that dependents can control when to accept breaking changes.

### Actual Behavior

This repository uses "prerelease" version numbers like `0.2.0-prerelease.YYYYMMDDhhmmss` which contain no information about breaking changes.

cwillisf:
### Resolves

- Resolves #1588
- Closes #1638, which was an earlier attempt at this.

### Proposed Changes

Configure semantic-release and supporting packages
Switch to `node:lts` image family on CircleCI

### Reason for Changes

This brings us into compliance with Semantic Versioning, and makes it more likely we'll catch compatibility problems with newer versions of Node.

### Test Coverage

The last few commits in this branch will show that everything is working for local development.
Building on CI will test most of the rest of the configuration changes.
The final deploy step will be tested once this is merged to `develop`...


.circleci/config.yml
```diff
@@ -1,110 +1,81 @@
...
       working_directory: ~/repo
       environment:
         NODE_ENV: production
   - &setup
-    name: "setup"
-    command: |
-      npm --production=false ci
-      mkdir ./test/results
+    node/install-packages:
+      override-ci-command: npm --production=false ci
```
cwillisf, line 18: Sorry for all the visual noise here. The changes in the `aliases` section are:

1. Convert all of these from just `name, command` to `run: {name: command}` so that step 2 is possible
2. Replace the contents of `setup` with the `node/install-packages` step from the `node` orb, instead of the previous `run` step.

delasare: Nice! Thanks for the explanation here.

.circleci/config.yml
```diff
@@ -1,7 +1,8 @@
 version: 2.1
 orbs:
-    node: circleci/node@5.0.2
     browser-tools: circleci/browser-tools@1.2.4
+    node: circleci/node@5.0.2
+    commitlint: conventional-changelog/commitlint@1.0.0
```
delasare: ❤️ the orbs!

delasare (approved):
Looks good to me! :shipit:

cwillisf:
(rebased to resolve `package-lock.json` conflict due to `scratch-l10n` update)

PR #2145, opened 2023-04-07
feat!: upgrade to React 18

aoneill01:
**Do not merge to `develop`!** 

*Creating a draft PR to prevent accidental merges. This is ready for review. Will likely be merged into a pre-release branch first.*

### Resolves

- [ENA-206](https://scratchfoundation.atlassian.net/browse/ENA-206)

### Proposed Changes

- Upgrade to React 18
- Update deprecated lifecycle methods, particularly converting `componentWillReceiveProps` to `componentDidUpdate`. This necessitated removing some `shouldComponentUpdate` methods, but they are mostly buttons that are not expensive to render.
- Upgrade to react-intl 5. I inspected the messages in scratch-l10n used in `scratch-paint`, and I do not believe any messages need to be updated due to [message syntax changes](https://formatjs.io/docs/react-intl/upgrade-guide-3x/#message-format-syntax-changes). 
- Change how hover selection works. Before it would store the item id in Redux state, then `selection-hoc.jsx` would remove the selection when it changes. Now it is possible for `selection-hoc.jsx` to miss a change if they come in too quickly because they can be batched and trigger a single re-render. To keep the same overall program structure, I updated the hover state to keep track of the hovered item id and an array of removed item ids. `selection-hoc.jsx` now processes that array of removed item ids. 
- Use `feat!:` to denote as a breaking change to increase the major version number
- The "coming soon" component and its dependencies was removed because it is not being used anywhere.

### Reason for Changes

This is part of an epic to upgrade several Scratch repositories to React 18.

### Test Coverage

- Updated / added tests for the hover selection reducer.
- Otherwise functionality should be unchanged, and the existing tests are valid.

[ENA-206]: https://scratchfoundation.atlassian.net/browse/ENA-206?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/containers/color-indicator.jsx
```diff
@@ -30,11 +31,10 @@ const makeColorIndicator = (label, isStroke) => {
...
             this._hasChanged = false;
         }
-        componentWillReceiveProps (newProps) {
-            const {colorModalVisible, onUpdateImage} = this.props;
-            if (colorModalVisible && !newProps.colorModalVisible) {
+        componentDidUpdate (prevProps) {
+            if (prevProps.colorModalVisible && !this.props.colorModalVisible) {
                 // Submit the new SVG, which also stores a single undo/redo action.
-                if (this._hasChanged) onUpdateImage();
+                if (this._hasChanged) this.props.onUpdateImage();
```
cwillisf, line 37: ~I'm not sure whether or not this is important, but if I'm reading this correctly the old code would call `onUpdateImage` from the previous props, where the new code calls it on the new props.~

I looked into where `onUpdateImage` comes from and I'm pretty convinced that (a) it never changes, and (b) if it ever did change, using the new value is probably more correct. So I think this change is fine :)

src/hocs/selection-hoc.jsx
```diff
@@ -13,23 +14,19 @@ const SelectionHOC = function (WrappedComponent) {
...
-                if (match) {
-                    match.remove();
-                }
+            const match = paper.project.getItem({
+                match: item => (item.id === itemId)
+            });
+            if (match) {
+                match.remove();
             }
+            this.props.clearRemovedItem(itemId);
```
cwillisf, line 29: Should this call to `clearRemovedItem` be inside the `if (match)`? It seems like it would only matter if it's possible to have the same item ID twice (or more) in `removedItemIds`. Can that happen? If so, would it be more correct to call `clearRemovedItem` twice (or more), or just once, on that ID?

src/reducers/hover.js
```diff
@@ -1,20 +1,45 @@
...
 const reducer = function (state, action) {
     if (typeof state === 'undefined') state = initialState;
     switch (action.type) {
-    case CHANGE_HOVERED:
+    case CHANGE_HOVERED: {
         if (typeof action.hoveredItemId === 'undefined') {
             log.warn(`Hovered item should not be set to undefined. Use null.`);
             return state;
-        } else if (typeof action.hoveredItemId === 'undefined' || isNaN(action.hoveredItemId)) {
+        } else if (isNaN(action.hoveredItemId)) {
```
cwillisf, line 17: 😎

cwillisf (approved):
I put in a couple of questions, but I have a feeling this is probably all good. Just checking my understanding, especially for the new hover code :)

Actually I'll go ahead and mark it "Approve" on the guess that none of my questions will amount to a real problem...

src/hocs/selection-hoc.jsx
aoneill01: Based on what I see in the redux state and in the code, I don't think we will see the same id twice or more in `removedItemIds`. Except for the rare situation that this new code is trying to address, there is either 0 or 1 items in `removedItemIds`. I'm leaning towards leaving it outside of the `if`. If there were ever multiples of the same id, the consequence is extra calls to `clearRemovedItem`. But it also protects `removedItemIds` from never being cleared out if some other bug resulted in an id that had already been removed from the project.

cwillisf:
Completed in #3334
Thank you SOOO much, @aoneill01, for your foundational work on this!

PR #2249, opened 2023-10-04
Move scratch-paint to GitHub Actions

delasare:
### Resolves

- Add .nvmrc
- create github actions workflow file.


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,64 @@
...
+          uses: peaceiris/actions-gh-pages@v3
+          with:
+            github_token: ${{ secrets.GITHUB_TOKEN }}
+            publish_dir: playground
+            full_commit_message: "Build for ${{ github.sha }} ${{ github.event.head_commit.message }}"
+        - name: Semantic Release
+          if: ${{ env.TRIGGER_DEPLOY == 'true'}}
+          env:
+            NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+            GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
skripted-io, line 61: Since the run command is not referring to these variables, these 3 lines can probably go?

skripted-io: Scrap that, this is actually the way it should be. Turns out "semantic-release" is picking up those variables in the background. 

In the workflows I worked on, I had to use "npm publish" instead of "npx semantic-release", and there you need to explicitly pass the token to the command.

package.json
```diff
@@ -7,13 +7,11 @@
...
     "build": "npm run clean && webpack --progress --colors --bail",
     "clean": "rimraf ./dist && mkdirp dist && rimraf ./playground && mkdirp playground",
-    "deploy": "touch playground/.nojekyll && gh-pages -t -d playground -m \"[skip ci] Build for $(git log --pretty=format:%H -n1)\"",
     "i18n:push": "tx-push-src scratch-editor paint-editor ./translations/en.json",
     "i18n:src": "rimraf ./translations/messages && babel src > tmp.js && rimraf tmp.js && ./scripts/build-i18n-source.js ./translations/messages/ ./translations/",
     "lint": "eslint . --ext .js,.jsx",
-    "prepare": "husky install",
     "start": "webpack-dev-server",
-    "test": "npm run lint && npm run unit && NODE_ENV=production npm run build",
+    "test": "npm run lint && npm run unit",
```
cwillisf, line 14: Thanks for catching this! I'm not sure why it was set up this way before.

package.json
```diff
@@ -7,13 +7,11 @@
   "scripts": {
     "build": "npm run clean && webpack --progress --colors --bail",
     "clean": "rimraf ./dist && mkdirp dist && rimraf ./playground && mkdirp playground",
-    "deploy": "touch playground/.nojekyll && gh-pages -t -d playground -m \"[skip ci] Build for $(git log --pretty=format:%H -n1)\"",
     "i18n:push": "tx-push-src scratch-editor paint-editor ./translations/en.json",
     "i18n:src": "rimraf ./translations/messages && babel src > tmp.js && rimraf tmp.js && ./scripts/build-i18n-source.js ./translations/messages/ ./translations/",
     "lint": "eslint . --ext .js,.jsx",
-    "prepare": "husky install",
```
cwillisf, line 14: Was this giving you trouble?

cwillisf (approved):
Looks good!

package.json
delasare: Yep. I was getting an error on install. 

Looking `wagoid/commitlint-github-action@v5`  it looks like there is a husky dependency included, so I was thinking it wasn't necessary to install it as a pre-install, install

cwillisf (approved):

PR #2257, opened 2023-10-17
ci: babel and other updates

delasare:
Resolves an issue with an older version of Babel that breaking the automated translation task:

1. The actual Babel changes are already on dev.
2. This branch is based on dev and addresses build and test issues.
3. ~Note that `npm ci` produces an error, but using `npm install` works.~
4. Use `canvas` package instead of `jest-canvas-mock`

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -38,9 +38,11 @@ jobs:
             EOF
         - name: Install NPM Dependencies
           run: |
-            npm install 
+            npm ci
             mkdir -p ./test/results
-        - name: Lint and Test
+        - name: Run Lint
+          run: npm run lint
+        - name: Run Tests
           run: npm run test
         - name: Build
           run: npm run build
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy.yml
```diff
@@ -3,8 +3,7 @@ name: Scratch Paint CI-CD
 on:
   pull_request: # Runs whenever a pull request is created or updated
   push: # Runs whenever a commit is pushed to the repository...
-    branches: [master, develop, beta, hotfix/*] # ...on any of these branches
-  workflow_dispatch: # Allows you to run this workflow manually from the Actions tab
+    # branches: [master, develop, beta, hotfix/*] # ...on any of these branches
```
cwillisf: Was this meant as a temporary change for testing purposes?

.github/workflows/deploy.yml
```diff
@@ -57,7 +58,7 @@ jobs:
         - name: Semantic Release
           if: ${{ env.TRIGGER_DEPLOY == 'true'}}
           env:
-            NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+            NPM_TOKEN: q${{ secrets.NPM_TOKEN }}
```
cwillisf: Stray `q`?

cwillisf (changes requested):
Looks like a couple unintended or temporary changes snuck in... the rest looks good, though :)

.github/workflows/deploy.yml
delasare: gah. yep that was for testing.

.github/workflows/deploy.yml
delasare: 🤦

cwillisf (approved):
🚀

PR #927, opened 2019-11-19
Use Paper.Color instead of CSS color strings

adroitwhiz:
### Resolves

Resolves #610

### Proposed Changes

This PR changes the paint editor to consistently use `Paper.Color` objects to store color instead of CSS color strings.

### Reason for Changes

This prevents unexpected side effects of HSV > RGB > HSV conversion, such as certain color channels being zeroed out when brightness or saturation are set to 0.

It also allows increased flexibility with regards to future color improvements, notably a "transparency" slider (#834).

Because this touches a huge portion of the codebase, I'd like to look over it with fresh eyes once some more time has passed. I'm submitting this as a draft in case you want to look over it too.

PR #2627, opened 2024-06-25
Opacity slider

### LINKED ISSUE #834: "Transparency" slider in color pickers
Feature request - in order to create all sorts of interesting effects (especially in combination with gradients), there should be a "transparency" slider in color pickers (fill in particular, though outline would be nice too), at the bottom (below brightness).

varunbiniwale:
### Resolves

#834 

### Proposed Changes

Adds a slider to the colour picker to change the opacity of colours + all the necessary plumbing to make it work
 
image

https://github.com/scratchfoundation/scratch-paint/assets/72911498/6a26663b-3135-4d76-8411-df19584a96e5

* Add the slider and prop to the color-picker component
* Add the handlers to the container component
* Change color functions to support an alpha value - some of this is hacky, I think scratch-paint could use some effort to work with color better - #927?
    * eg. `parse-color` from npm is unmaintained, and returns inconsistently. Hex colours like #00ffb9 have an alpha of 1, and rgba output ranges from 0-1, but for hex with the alpha part (eg. #00ffb9de) the scale is from 0-255
    * I made an attempt to work with this, but it could definitely be better
* Adjust eyedropper to support alpha values (this has the side effect of making picking colors from between a gradient from opaque to transparent much more useful)

https://github.com/scratchfoundation/scratch-paint/assets/72911498/2b5c8a10-937c-4518-90aa-43bac3b260ff

### Reason for Changes

Adds functionality to the paint editor that was previously difficult, introduces new users to the concept of colour transparency, etc.

### Test Coverage

TODO


varunbiniwale:
Ideally this would work in both the bitmap and vector editors. Currently, certain tools in the bitmap editor (eg. paint brush and circle tool) don't support alpha.

I can either work on implementing those (increasing the surface area of changes in this PR), or limit the new opacity slider to the vector editor. I'm looking into what kinds of changes need to be done to the bitmap tools to support the alpha channel.

rosics-code:
ST see this !!

PR #3334, opened 2025-07-23
React 18 integration

KManolov3:
### Resolves

https://scratchfoundation.atlassian.net/browse/UEPR-63

### Proposed Changes

Upgrade scratch-paint to use react v18

### Reason for Changes

React v16 is getting increasingly deprecated and is not supported by many libraries.


cwillisf (approved):
🚀

PR #3515, opened 2026-04-04
fix(bitmap): Prevent crash when rotating selection with negative bounds

### LINKED ISSUE #1129: Part of the drawing disappears after rotating it in the bitmap editor
### Expected Behavior
Rotating a selection does not delete it

### Actual Behavior
After I rotated in bitmap mode, the selection doesn’t appear on the stage
Screen Shot 2020-06-18 at 12 18 41

And it disappeared in the paint editor when I clicked away from the selection
Screen Shot 2020-06-18 at 12 18 49

Console:
Screen Shot 2020-06-18 at 12 19 31

### Steps to Reproduce
Eric also saw this once during the bug hunt but neither of us could repro

Windows Chrome and Edge

aryanjsingh:
## Summary

This fixes a bug where rotating a selection in the bitmap editor could cause it to disappear. The issue was caused by `getHitBounds` being called with a non-normalized rectangle (negative width/height), leading to an error in `getImageData`. The fix normalizes the bounds within `getHitBounds` to ensure dimensions are always positive, preventing the crash.

## Changes

- **src/helper/bitmap.js**: In `src/helper/bitmap.js`, the `getHitBounds` function is modified to normalize the bounds rectangle before calling `raster.getImageData`. This prevents a `DOMException` when `getImageData` is called with negative width or height, which can occur when a selection is rotated. A new `normalizedBounds` variable is created using `new paper.Rectangle(bounds.topLeft, bounds.bottomRight)`, and this is used for all subsequent calculations within the function.

## Related Issue

Closes #1129

### DIFF
--- a/src/helper/bitmap.js
+++ b/src/helper/bitmap.js
@@ -356,8 +356,10 @@ const columnBlank_ = function (imageData, width, x, top, bottom) {
  */
 const getHitBounds = function (raster, rect) {
     const bounds = rect || raster.bounds;
-    const width = bounds.width;
-    const imageData = raster.getImageData(bounds);
+    // Normalize bounds to handle rectangles with negative width/height.
+    const normalizedBounds = new paper.Rectangle(bounds.topLeft, bounds.bottomRight);
+    const width = normalizedBounds.width;
+    const imageData = raster.getImageData(normalizedBounds);
     let top = 0;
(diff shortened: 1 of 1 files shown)


aryanjsingh:
I have read the CLA Document and I hereby sign the CLA

